PR #57, opened 2015-02-03
Add decimal precision and scale to HS2 cursor description

wesm:
Not being a Thrift expert, there might be a less hackish way to get at the precision/scale values in the RPC request.


laserson:
Nah, I think you got it :) Thanks, @wesm.  I'll create a patch release for this later this week.

wesm:
Sweet, thanks

PR #66, opened 2015-04-03
Query parameters

tjeb:
Even though PEP 249 only requires cursor.execute() to support one parameter style (the one specified in the module's paramstyle value, in this case pyformat), I think it would be nice to also support some of the others, similar to what the sqlite3 module does.

This branchs adds support for named style (first commit), question mark style and numeric style (second commit). The third commit adds some exception for commonly used errors (though I left the pyformat replacement as it is, e.g. it would raise a KeyError for a nonexistent parameter, as it already did). Final commit is a slight documentation addition to the specialized tokenizer (added in the first commit since the python tokenizers don't really appear to fit the use-case).

The default (and paramstyle value) is still pyformat.

### DIFF
--- a/impala/dbapi/interface.py
+++ b/impala/dbapi/interface.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import re
 from six import reraise
 
 from impala.util import _escape
@@ -164,9 +165,55 @@ class Cursor(object):
         if exc_type is not None:
             reraise(exc_type, exc_val, exc_tb)
 
+def _replace_numeric_markers(operation, string_parameters):
+    """
+    Replaces qname and numeric markers in the given operation, from
+    the string_parameters list.
+    Raises ProgrammingError on wrong number of parameters or bindings
+    when using qmark. There is no error checking on numeric parameters.
+    """
+    # replace qmark parameters
+    param_count = len(string_parameters)
+    qmark_index = 0
+    while operation.find('?') > -1:
+        if qmark_index < param_count:
+            operation = operation.replace('?',
+                                          string_parameters[qmark_index],
+                                          1)
+            qmark_index += 1
+        else:
+            raise ProgrammingError("Incorrect number of bindings "+
+                "supplied. The current statement uses "
+                "%d or more, and there are %d supplied." %
+                (qmark_index+1, param_count))
+    if qmark_index != 0 and qmark_index != param_count:
+        raise ProgrammingError("Incorrect number of bindings "+
+            "supplied. The current statement uses "
+            "%d or more, and there are %d supplied." %
+            (qmark_index+1, param_count))
 
-def _bind_parameters(operation, parameters):
-    # inspired by MySQL Python Connector (conversion.py)
+    # replace numbered parameters
+    # Go through them backwards so smaller numbers don't replace
+    # parts of larger ones
+    for index in range(param_count, 0, -1):
+        operation = operation.replace(':' + str(index),
+                                      string_parameters[index-1])
+    return operation
+
+def _bind_parameters_list(operation, parameters):
+    string_parameters = []
+    for value in parameters:
+        if value is None:
+            string_parameters.append('NULL')
+        elif isinstance(value, basestring):
+            string_parameters.append("'" + _escape(value) + "'")
+        else:
+            string_parameters.append(str(value))
+
+    # replace qmark and numeric parameters
+    return _replace_numeric_markers(operation, string_parameters)
+
+def _bind_parameters_dict(operation, parameters):
     string_parameters = {}
     for (name, value) in parameters.iteritems():
         if value is None:
@@ -175,4 +222,20 @@ def _bind_parameters(operation, parameters):
             string_parameters[name] = "'" + _escape(value) + "'"
         else:
             string_parameters[name] = str(value)
+
+    # replace named parameters by their pyformat equivalents
+    operation = re.sub(":([^\d\W]\w*)", "%(\g<1>)s", operation)
+
+    # replace pyformat parameters
     return operation % string_parameters
+
+def _bind_parameters(operation, parameters):
+    # If parameters is a list, assume either qmark or numeric
+    # format. If not, assume either named or pyformat parameters
+    if isinstance(parameters, (list, tuple)):
+        return _bind_parameters_list(operation, parameters)
+    elif isinstance(parameters, dict):
+        return _bind_parameters_dict(operation, parameters)
+    else:
+        raise ProgrammingError("Query parameters argument should be a "+
+                               "list, tuple, or dict object")
--- a/impala/tests/test_dbapi_query_parameters.py
+++ b/impala/tests/test_dbapi_query_parameters.py
@@ -1,0 +1,216 @@
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Additional tests specific for query parameters
+#
+
+from __future__ import absolute_import
+import sys
+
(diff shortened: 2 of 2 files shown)


laserson:
Wow, this is great, @tjeb!  Let me take a look at it and also check with the impala people (e.g., @wesm or @caseyching).

laserson:
and/or @ishaan

caseyching:
Nice! Doing a simple string replace instead of trying to parse a query might be better though. Something like https://github.com/mysql/mysql-connector-python/blob/master/lib/mysql/connector/cursor.py#L492 ? The parsing is missing a lot of cases, for example I think "foo = (:bar)" won't parse correctly? Also some tests should be added for when the replacement value itself contains a quote.

tjeb:
Good point. I did consider it at first, but I was worried about quoted parts in queries that look like keywords ("where foo = ':bar') and it's nice for the programmer if there is an exception when a keyword is used that is not in the parameters (which I think that example misses, ending up with the relatively unhelpful 'parse error' exception). But this does indeed miss a number of cases, which would probably require a full parser (ending up with the operation being parsed twice...). And regexes don't tend to fit these types of things well either.

Perhaps a simpler split into quotes/nonquotes (similar to shlex.split()), then a string replace on the unquoted parts, then a quick scan for missed keywords? It's also possible that I am overthinking this, btw :)

wesm:
Two questions:

- Any concerns about unicode problems with this approach?
- Is this vulnerable to SQL injection (IIRC one of the benefits of the `?` DBAPI parameter approach is that it protects against SQL injection attack issues. Someone ought to look into that though)?

caseyching:
About string literals that contain parameter names, skipping over them while doing a search/replace on params sounds great. I think it's also fine to require the query author to use different parameter names so there is no conflict. "where foo = ':bar' and bar = :bar" could be rewritten as "where foo = :foo and ..." or "where foo = ':bar' and bar = :baz".

tjeb:
Ok, I've removed the custom tokenizer, and simply use string.replace(). This also removed much of the new error checking though, which would then require another pass through the operation (with added chances of failure :) ).

Regarding Unicode and SQL Injection; this branch only touches the _bind_parameters() call, so it should at the very most be bug-compatible, but I haven't checked if mixing unicode and non-unicode could be a problem. String arguments go through the same _escape() call as the pyformat replacement used.

impala/dbapi/interface.py
```diff
@@ -156,8 +156,70 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+    for value in parameters:
+        if value is None:
+            string_parameters.append('NULL')
+        elif isinstance(value, basestring):
+            string_parameters.append("'" + _escape(value) + "'")
+        else:
+            string_parameters.append(str(value))
 
-def _bind_parameters(operation, parameters):
+    # paramstyle is pyformat, but support named as well
```
laserson: I think this comment is from a copy/paste

impala/dbapi/interface.py
```diff
@@ -156,8 +156,70 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+    # replace numbered parameters
+    # Go through them backwards so smaller numbers don't replace
+    # parts of larger ones
+    for index in range(param_count, 0, -1):
+        operation = operation.replace(':' + str(index),
+                                      string_parameters[index-1])
+    return operation
+
+def _bind_parameters_list(operation, parameters):
+    # inspired by MySQL Python Connector (conversion.py)
```
laserson: is this still true for this fn?

impala/dbapi/interface.py
```diff
@@ -167,4 +229,17 @@ def _bind_parameters(operation, parameters):
...
+    
+    # but default to pyformat
     return operation % string_parameters
+
+def _bind_parameters(operation, parameters):
+    # If parameters is a list, assume either qmark or numeric
+    # format. If not, assume either named or pyformat parameters
+    if isinstance(parameters, (list, tuple)):
+        return _bind_parameters_list(operation, parameters)
+    else:
```
laserson: Could this catch a generator object by accident?  Would anyone ever do that?

laserson:
Looks good, and big +1 for adding tests :)  Could you squash this into a single commit?

laserson:
Also, could you rebase this on master?

tjeb:
Updated those comment lines, and squashed and rebased on master.

impala/dbapi/interface.py
```diff
@@ -164,9 +164,69 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+    """
+    Checks if there are any named markers as described in PEP 249,
+    if so replaces them from the given string_parameters.
+    Does *not* raise an error for unused parameters, or for parameters
+    that are not found.
+    """
+    # Iterate through the keywords, longest first, so partially
+    # overlapping ones don't replace the wrong value (:foo and :foobar)
+    for key in sorted(string_parameters, reverse=True):
+        operation = operation.replace(':' + key, string_parameters[key])
```
laserson: What happens if

```python
string_parameters = {
    'foo': ':bar',
    'bar': 'bummer'}
```

laserson:
One thing that I think makes things more predictable and seems more compliant with PEP 249 would be if we explicitly check what `paramstyle` the user wants to use.  The global `paramstyle` value can default to `'pyformat'`, but the user can change this value to one of the other ones.  Then the proper parameter binding code can be selected explicitly, and there are fewer surprises.

Instead of manipulating the global variable, this could also be an extra parameter passed to the execution code that defaults to pyformat.

impala/dbapi/interface.py
tjeb: hmm. yes, in this version that will be rather unpredictable (depening on key size it will be replaced again, or not). Which suggest it does indeed need something of a scanner instead of string.replace(). hmm.

tjeb:
PEP 249 states that paramstyle is a constant though. I think the specification misses an opportunity here (imho it should either be a list of supported types, or it should have a way to set it, preferably with an error right there if the style is not supported). I think that both making it non-constant or passing it to execute deviates from PEP 249 further than implicitely supporting other styles.

impala/dbapi/interface.py
laserson: Would it be possible/practical to instead convert "named" to "pyformat" via regex?  So replace each `":bar"` with `"%(bar)s"` or whatever?

tjeb: damn how did I not think of that. Yes, that should work, and have the same behaviour as the pyformat case, i.e. a KeyError for missing parameters.

tjeb:
Pushed a commit that changes the named parameters to pyformat ones. Separately for now, in case there's other comments or requests (or people disagree about my view of pep249's paramstyle), but happy to rebase and squash again if people think this is ready.

impala/dbapi/interface.py
```diff
@@ -20,6 +20,8 @@
                           IntegrityError, DataError, NotSupportedError)
 
 
+import re
```
laserson: nit: can you move this import to be above `from six import reraise`

impala/dbapi/interface.py
```diff
@@ -175,4 +237,17 @@ def _bind_parameters(operation, parameters):
             string_parameters[name] = "'" + _escape(value) + "'"
         else:
             string_parameters[name] = str(value)
+
+    # replace named parameters by their pyformat equivalents
+    operation = re.sub(":([a-zA-Z]+)", "%(\g<1>)s", operation)
```
laserson: I don't know whether there is a "spec" for the "named" mechanism, but can the identifiers include digits and underscores?  cf. http://stackoverflow.com/a/10134719/510187

impala/dbapi/interface.py
```diff
@@ -164,9 +166,69 @@ def __exit__(self, exc_type, exc_val, exc_tb):
         if exc_type is not None:
             reraise(exc_type, exc_val, exc_tb)
 
+def _replace_named_markers(operation, string_parameters):
```
laserson: so this fn can be deleted now, no?

impala/dbapi/interface.py
laserson: Put another way, should we test `elif isinstance(parameters, dict)` and then raise an exception on the `else` branch?

impala/dbapi/interface.py
tjeb: couldn't find one, but [^\d\W]\w* seems to make sense.

laserson:
cool, rebase/squash and I'll merge.

tjeb:
awesome, thanks!

laserson:
Thanks, @tjeb, great contribution!

PR #67, opened 2015-04-07
Fix thrift_sasl.py to properly support SASL QOP

mjacobs:
See Impala IMPALA-1565, commit c3e0a5033ba6795d6bb422e593d6170633c136ec


laserson:
thx, @mjacobs!

PR #70, opened 2015-04-17
adding head() for BigDataFrame

cgc17:
returns a pandas DataFrame using impala.util.as_pandas()


laserson:
thanks @cgc17!  Does the `.take()` method not suit your needs?  (Or not work?)

cgc17:
Whoops! @laserson I didn't even catch that method as I was browsing through the code. I'm so used to using `.head()` with pandas that I must have read right past it. I'll close this for now and if I have any problems with `.take()`, I'll let you know with an issue. Thanks!

Issue #27, opened 2014-09-19
Support Python 3


mrocklin:
I'm happy to help with this if testing can be made convenient.

laserson:
Fixed by #74.

PR #74, opened 2015-06-03
Support Python 3 (including Kerberos)

laserson:
Some of the main changes:

* Updated syntax to be Python 2/3 compatible
* Use `thriftpy` with Python 3 instead of `thrift`.
* `thrift_sasl.py` has been refactored a bit, including a key change to `.read()` to make it work with Kerberos
* `setup.py` now installs the proper code base depending on whether it's invoked with Python 2 or 3.  (How is packaging handled in this situation?)

cc @jwills @wesm 

Wes: I've never ported a package to Python 3, so let me know if anything here is idiotic.

Things that are outstanding:

* What is the proper way to deal with packaging?
* SSL is still not supported


impala/_rpc/beeswax.py
```diff
@@ -12,17 +12,58 @@
...
+    beeswax = load(os.path.join(thrift_dir, 'beeswax.thrift'),
+                   include_dirs=[thrift_dir])
+    sys.modules[ExecStats.__name__] = ExecStats
+    sys.modules[Status.__name__] = Status
+    sys.modules[ImpalaService.__name__] = ImpalaService
+    sys.modules[beeswax.__name__] = beeswax
+    from ExecStats import TExecStats
+    from Status import TStatus, TStatusCode
+    from ImpalaService import ImpalaService
+    import beeswax as BeeswaxService
```
wesm: yikes, is there no canonical way to do this? perhaps it would be better to factor this into a thrift_api compatibility module so that you don't have to rely on such black magic

laserson: Yeah, it's extra ugly.  I'll move it to a separate module.

impala/_rpc/hiveserver2.py
```diff
@@ -117,8 +163,11 @@ def wrapper(*args, **kwargs):
         # get the thrift transport
         if 'service' in kwargs:
             transport = kwargs['service']._iprot.trans
-        elif (len(args) > 0 and
-                  isinstance(args[0], ImpalaHiveServer2Service.Client)):
+        elif (six.PY2 and len(args) > 0 and
+                isinstance(args[0], ImpalaHiveServer2Service.Client)):
+            transport = args[0]._iprot.trans
+        elif (six.PY3 and len(args) > 0 and
+                isinstance(args[0], TClient)):
```
wesm: pep8

impala/thrift_sasl.py
```diff
@@ -20,20 +17,46 @@
...
+import sasl
+
+
+PY2 = sys.version_info[0] == 2
+PY3 = sys.version_info[0] == 3
+
+
+if PY2:
+  from cStringIO import StringIO
+  from thrift.transport.TTransport import (
```
wesm: yikes mixed indentation within the same package =(

wesm:
LGTM outside the thrift import black magic. Good stuff

impala/thrift_sasl.py
laserson: Yeah, this is the only exceptional case, bc this file has been copied around from multiple repos.  I'll likely contribute these changes back to the main impala repo and Hue.

laserson:
The Thrift API has now been factored out into the `_thrift_api` subpackage.

laserson:
Fixes #27.

PR #88, opened 2015-07-23
Replace dependency on sasl with pure-sasl

nonsleepr:
Installation of python-sasl is complicated on Windows, while pure-sasl is pure.


laserson:
Thanks for this!  I'm a fan of things that make it easier for people to use.  However, AFAIK the rest of the Hadoop stack uses the `python-sasl` package, which I am wary of mucking with.  My preference would be to have a `sasl_compat.py` module which defaults to `python-sasl` and falls back on `pure-sasl` while emitting a warning.  You can then abstract over the two modules.

cc @caseyching @ishaan

impala/tests/test_dbapi_connect.py
```diff
@@ -0,0 +1,115 @@
+# Copyright 2013 Cloudera Inc.
```
laserson: These tests look great!  Though I need to see if I have access to internal test clusters for all the options.

impala/thrift_sasl.py
```diff
@@ -24,7 +24,7 @@
 import sys
 import struct
 
-import sasl
+#import sasl
```
laserson: Just delete this if it's unused.

laserson:
@nonsleepr Just to check, are you going to resubmit a modified PR?

impala/tests/test_dbapi_connect.py
nonsleepr: Only 3 of 8 are successful though with default Cloudera VM settings.

nonsleepr:
@laserson Resubmitted as #91.

PR #89, opened 2015-07-24
Expose async execute ops for HS2

caseyching:
With this change queries can be executed asynchronously using HS2.

Example usage:
  cursor.execute_async("SELECT 1")
  while cursor.is_executing():
    sleep(1)
  if cursor.execution_failed():
    # Handle error
  cursor.fetchall()

This allows for an easy way to enforce a query timeout without the need
for threading.


laserson:
Awesome, @caseyching!  Will merge tomorrow if not later today.

laserson:
Manually merged in c5dfcff.  Great feature, @caseyching!

PR #91, opened 2015-07-24
Compatibility layer to allow use of either python-sasl or pure-sasl

nonsleepr:


impala/sasl_compat.py
```diff
@@ -0,0 +1,72 @@
+try:
+    import sasl
+
+    def build_sasl_factory(host, use_ldap, username, password, service):
```
nonsleepr: Factory for sasl_factory is probably overkill, but it allows to have all sasl-related code in one file.

laserson:
Looks good!  There is some poor timing, but I want to merge #93 first, which will require some more changes in this PR.  Sorry about all the confusion.  Also, in case you didn't know, there is no need to close the PR to make changes.  Just feel free to push more commits to it (or force push if you rebase).  We'll just squash them at the end anyway.

laserson:
Actually, I discussed internally, and we probably don't want to have the module fall back on pure-sasl, as it can be a liability with clients.  Let's make it so that there is a flag that you have to explicitly request the pure-sasl module if you want it.  I think we'd need to put in place some extensive testing of the pure-sasl module before we feel comfortable falling back on it.

wesm:
I know it's been a while but any chance of rebasing this on master?

nonsleepr:
Yes, will do that in next couple days.

PR #92, opened 2015-07-24
*_exists functions ignore case

nonsleepr:
Fixes issue where database_exists and table_exists functions work only with lowercase names.


laserson:
Good catch, @nonsleepr!  Thanks!

PR #93, opened 2015-07-24
Add support for configuring authentication mechanisms (required for H…

lskuff:
…ive)

Hive's default transport is PLAIN SASL where Impala does not use SASL. Impala and Impyla
support supported PLAIN SASL, but only for LDAP connections. This change updates Impyla
to allow the authentication mechanism to be specified as part of the connect() call.

For example, connections to Hive can be made using:
conn = connect(host='host', port=10000, auth_mechanism='PLAIN')

The valid values for auth_mechanism are:
- LDAP - Uses plain SASL, but there are some differences in how passwords are handled.
- PLAIN - Authenticate using plain SASL
- GSSAPI - Authenticate using SASL with GSSAPI (Kerberos)
- NOSASL - Don't use sasl.

auth_mechanism is optional and if it is not specifieid the default stays will be the same
as it is today - NOSASL. Support for the use_kerberos / use_ldap flags also is allowed
in which case the mechanism will be set to the appropriate value.

As part of this change there was a bunch of duplication between Beeswax and HS2 so this
fixes that problem by putting the common code in a thrift_util module.

### DIFF
--- a/impala/_rpc/beeswax.py
+++ b/impala/_rpc/beeswax.py
@@ -21,8 +21,9 @@ from six.moves import map
 from six.moves import range
 
 from impala.error import RPCError, QueryStateError, DisconnectedError
+from impala._rpc import thrift_util
 from impala._thrift_api.beeswax import (
-    TSocket, TBufferedTransport, TTransportException, TBinaryProtocol,
+    TTransportException, TBinaryProtocol,
     TApplicationException, BeeswaxService, ImpalaService, TStatus, TStatusCode,
     TExecStats, ThriftClient)
 
@@ -169,28 +170,16 @@ def build_summary_table(summary, idx, is_fragment_root, indent_level, output):
     return idx
 
 
-def _get_socket(host, port, use_ssl, ca_cert):
-    # based on the Impala shell impl
-    if use_ssl:
-        from thrift.transport.TSSLSocket import TSSLSocket
-        if ca_cert is None:
-            return TSSLSocket(host, port, validate=False)
-        else:
-            return TSSLSocket(host, port, validate=True, ca_certs=ca_cert)
-    else:
-        return TSocket(host, port)
-
-
 def connect_to_impala(host, port, timeout=45, use_ssl=False, ca_cert=None,
-                      use_ldap=False, ldap_user=None, ldap_password=None,
-                      use_kerberos=False, kerberos_service_name='impala'):
-    sock = _get_socket(host, port, use_ssl, ca_cert)
+                      username=None, password=None, kerberos_service_name='impala',
+                      auth_mechanism=None):
+    sock = thrift_util.get_socket(host, port, use_ssl, ca_cert)
     if six.PY2:
         sock.setTimeout(timeout * 1000.)
     elif six.PY3:
         sock.set_timeout(timeout * 1000.)
-    transport = _get_transport(sock, host, use_ldap, ldap_user, ldap_password,
-                               use_kerberos, kerberos_service_name)
+    transport = thrift_util.get_transport(sock, host, kerberos_service_name, auth_mechanism,
+        username, password)
     transport.open()
     protocol = TBinaryProtocol(transport)
     if six.PY2:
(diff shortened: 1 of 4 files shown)


impala/_rpc/beeswax.py
```diff
@@ -21,8 +21,9 @@
 from six.moves import range
 
 from impala.error import RPCError, QueryStateError, DisconnectedError
+from impala._rpc import thrift_util
```
laserson: Nit: for consistent style, maybe change to `from impala._rpc.thrift_util import ...`

impala/_rpc/hiveserver2.py
```diff
@@ -35,8 +35,9 @@
 from six.moves import range
 
 from impala.error import HiveServer2Error
+from impala._rpc import thrift_util
```
laserson: Same nit

impala/_rpc/thrift_util.py
```diff
@@ -0,0 +1,68 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Thrift utility functions
+
+from thrift.transport.TSocket import TSocket
+from thrift.transport.TTransport import TBufferedTransport
```
laserson: These classes actually need to be imported from the `_thrift_api` submodule.  Looking at this, it appears there's some unnecessary duplication there as well.  Perhaps this whole `thrift_util.py` file can be moved into `_thrift_api/__init__.py`?  And `TSocket` and `TBufferedTransport` can be moved into the common `__init__.py`?

impala/_rpc/thrift_util.py
```diff
@@ -0,0 +1,68 @@
...
+            return TSSLSocket(host, port, validate=True, ca_certs=ca_cert)
+    else:
+        return TSocket(host, port)
+
+
+def get_transport(socket, host, kerberos_service_name, auth_mechanism="NOSASL",
+                  user=None, password=None):
+    """Creates a new Thrift Transport using the specified auth_mechanism.
+    Supported auth_mechanisms are:
+    - None or NOSASL - returns simple buffered transport (default)
```
laserson: Nit: add single quotes around `NOSASL` and the other ones to make clear that you pass a string.

impala/dbapi/__init__.py
```diff
@@ -33,21 +33,42 @@
...
 def connect(host='localhost', port=21050, protocol='hiveserver2',
             database=None, timeout=45, use_ssl=False, ca_cert=None,
             use_ldap=False, ldap_user=None, ldap_password=None,
-            use_kerberos=False, kerberos_service_name='impala'):
+            use_kerberos=False, kerberos_service_name='impala',
+            auth_mechanism=None):
+
+    # Supported authentication mechanisms
+    auth_mechanisms = ['NOSASL', 'PLAIN', 'GSSAPI', 'LDAP']
+    if use_kerberos:
```
laserson: I like how you eliminated the `use_kerberos` option in the other functions.  Do you think it's safe to do so here?  Or should we keep it for backwards compatibility?  Perhaps we check whether `use_kerberos=True`, and if so, emit a deprecation message and manually set `auth_mechanism='GSSAPI'?`

impala/dbapi/__init__.py
```diff
@@ -33,21 +33,42 @@
...
+        if auth_mechanism and auth_mechanism.upper() != 'GSSAPI':
+            raise ValueError("Kerberos requires authentication mechanism 'GSSAPI'")
+        else:
+            auth_mechanism = 'GSSAPI'
+
+    if use_ldap:
+        if auth_mechanism and auth_mechanism.upper() != 'LDAP':
+            raise ValueError("LDAP requires authentication mechanism 'LDAP'")
```
laserson: Maybe `InterfaceError` or `ProgrammingError`?  (from `impala.error`)

impala/dbapi/__init__.py
```diff
@@ -33,21 +33,42 @@
...
+        else:
+            auth_mechanism = 'LDAP'
+
+    # If not specified, authentication mechanism defaults to NOSASL
+    if not auth_mechanism: auth_mechanism = 'NOSASL'
```
laserson: Nit: make it two lines (per PEP8).

impala/dbapi/__init__.py
```diff
@@ -33,21 +33,42 @@
...
-
 def connect(host='localhost', port=21050, protocol='hiveserver2',
             database=None, timeout=45, use_ssl=False, ca_cert=None,
             use_ldap=False, ldap_user=None, ldap_password=None,
-            use_kerberos=False, kerberos_service_name='impala'):
+            use_kerberos=False, kerberos_service_name='impala',
+            auth_mechanism=None):
+
+    # Supported authentication mechanisms
+    auth_mechanisms = ['NOSASL', 'PLAIN', 'GSSAPI', 'LDAP']
```
laserson: This feels like it should be a global variable in the module, no?

laserson:
Thanks, @lskuff!  Generally looks great.

laserson:
Picked up by @nonsleepr in #97.  Thanks for starting this, @lskuff!

PR #96, opened 2015-07-25
executemany has an option for 'bulk'-insert on Impala

nonsleepr:
Impala supports multiple row inserts. This pull request adds option to use this feature.

### DIFF
--- a/impala/dbapi/hiveserver2.py
+++ b/impala/dbapi/hiveserver2.py
@@ -201,10 +201,10 @@ class HiveServer2Cursor(Cursor):
             return 0.5
         return 1.0
 
-    def _exec_single_query_insert(self, operation, seq_of_parameters):
+    def _rewrite_as_bulk_insert(self, operation, seq_of_parameters):
         """
         Builds and executes one big INSERT...VALUES statement instead of sending
-        sequency of one-row inserts.
+        sequence of one-row inserts.
         Assumes that `operation` is INSERT...VALUES statement.
         Returns `True` if operation was executed and `False` otherwise.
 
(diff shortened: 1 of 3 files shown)


laserson:
I'm not sure I understand what this does.  Could you explain it in a bit more detail?

nonsleepr:
The idea here is to INSERT data in big chunks instead of doing it row-by-row.

laserson:
Ah, I was unfamiliar with `executemany`'s use as a method to insert rows.  Is this something you're doing often with your Impala cluster?

That said, I'm generally uneasy about having impyla rewrite people's queries, and anyway, using an `INSERT` statement is generally discouraged, as opposed to writing files to HDFS and registering the tables with `CREATE` statements.  It's only really implemented to comply with PEP 249.  My preference would be for you to send data to Impala using Ibis.  Thoughts?

nonsleepr:
In my project, I'm using **impyla** as one of several database drivers which could be accessed via DB API 2.0. Users should be able to upload/insert small datasets/tables (probably 100 rows max) into the database. Right now this will produce hundreds of files in HDFS while this PR allows to avoid it.

According to PEP 249:

> [`.executemany(...)`](https://www.python.org/dev/peps/pep-0249/#executemany)
> ...
> Modules are free to implement this method using multiple calls to the `.execute()` method **or by using array operations to have the database process the sequence as a whole in one call**.

[Impala docs](http://www.cloudera.com/content/cloudera/en/documentation/cloudera-impala/v2-0-x/topics/impala_insert.html#values_unique_1) also give following recommendations on `INSERT...VALUES`:

> If you do run `INSERT ... VALUES` operations to load data into a staging table as one stage in an ETL pipeline, **include multiple row values if possible within each `VALUES` clause**, and use a separate database to make cleanup easier if the operation does produce many tiny files.

I didn't have a chance to look at **ibis** until now. Interestingly it uses **impyla** and **hdfs** (which my project is based on) and bunch of other packages underneath. One of the goals of my project is to make it *lean* and preferably *pure* (and that's #91 is for).

laserson:
Ok, as long as the default is the same, shouldn't be a problem.  I'll make some additional comments for changes as well.

impala/dbapi/interface.py
```diff
@@ -25,6 +25,8 @@
                           IntegrityError, DataError, NotSupportedError)
 
 
+RE_INSERT_VALUES = re.compile(r"(.*\binsert\b.*\bvalues\b\s*)(\(.*\))\s*;?\s*", flags=re.IGNORECASE)
+
```
laserson: Nit: insert addl line per PEP8

impala/dbapi/hiveserver2.py
```diff
@@ -201,13 +201,49 @@ def _get_sleep_interval(self, start_time):
...
+                self._last_operation_string = query_left + ", ".join(bound_params)
+                self._last_operation_handle = rpc.execute_statement(
+                    self.service, self.session_handle, self._last_operation_string,
+                    None)
+
+            self._execute_sync(op)
+            return True
+        return False
+
+    def executemany(self, operation, seq_of_parameters, single_query_insert=False):
```
laserson: Perhaps change the name of the new flag to `rewrite_inserts_single_query`, or something like that?  I think it would be more clear as to exactly what's happening.

impala/dbapi/hiveserver2.py
```diff
@@ -201,13 +201,49 @@ def _get_sleep_interval(self, start_time):
             return 0.5
         return 1.0
 
-    def executemany(self, operation, seq_of_parameters):
+    def _exec_single_query_insert(self, operation, seq_of_parameters):
```
laserson: Instead of `_exec_single_query_insert()`, could you have a method called `_rewrite_inserts_single_query()` or something like that, which takes in `operation` and `seq_of_parameters` and returns the combined single SQL query string?

laserson: And returns `None` if it fails.

impala/dbapi/hiveserver2.py
```diff
@@ -201,13 +201,49 @@ def _get_sleep_interval(self, start_time):
...
+
+        # If condition above holds, we either don't want to build single insert
+        # query or were not able to build it.
+        if not single_query_insert or \
+           not self._exec_single_query_insert(operation, seq_of_parameters):
+            for parameters in seq_of_parameters:
+                self.execute(operation, parameters)
+                if self.has_result_set:
+                    raise ProgrammingError("Operations that have result sets are "
+                                           "not allowed with executemany.")
```
laserson: I find the logic here a big confusing.  I'd prefer something more like

```python
single_query = None
if rewrite_inserts_single_query:
    single_query = self. _exec_single_query_insert(...)
if single_query:
    self.execute(single_query)
else:
    for parameters in seq_of_parameters:
        ...
```

Gatsby-Lee:
@laserson Can you merge this change into master?

laserson:
I'm no longer involved with this project. Try a more recent committer.

timarmstrong:
Sorry this got neglected.

This is interesting but I'm unsure if the interface is quite right. Is there a reason not to do the rewrite transparently?

darklord1807:
Hi I achieved the same thing by this #460 and it is working well. 

Using 

  allToBeInserted.to_sql('xx', engine, if_exists='append', index=False,chunksize=2000,method='multi')

csringhofer:
@nonsleepr 
Closing this it was not touched for a long time and a last comment mentions, bulk inserts can be done using sqlalchemy (#460 ).
Feel free to reopen it - I can look at it again once conflicts are resolved.

PR #105, opened 2015-08-11
Add support for configuring authentication mechanisms

lskuff:
Hive's default transport is PLAIN SASL where Impala does not use SASL. Impala and Impyla
support supported PLAIN SASL, but only for LDAP connections. This change updates Impyla
to allow the authentication mechanism to be specified as part of the connect() call.

For example, connections to Hive can be made using:
conn = connect(host='host', port=10000, auth_mechanism='PLAIN')

The valid values for auth_mechanism are:

LDAP - Uses plain SASL, but there are some differences in how passwords are handled.
PLAIN - Authenticate using plain SASL
GSSAPI - Authenticate using SASL with GSSAPI (Kerberos)
NOSASL - Don't use sasl.
auth_mechanism is optional and if it is not specifieid the default stays will be the same
as it is today - NOSASL. Support for the use_kerberos / use_ldap flags also is allowed
in which case the mechanism will be set to the appropriate value.

As part of this change there was a bunch of duplication between Beeswax and HS2 so this
fixes that problem by putting the common code in a thrift_util module.

### DIFF
--- a/impala/_rpc/beeswax.py
+++ b/impala/_rpc/beeswax.py
@@ -21,8 +21,9 @@ from six.moves import map
 from six.moves import range
 
 from impala.error import RPCError, QueryStateError, DisconnectedError
+from impala._thrift_api import (
+    get_socket, get_transport, TTransportException, TBinaryProtocol)
 from impala._thrift_api.beeswax import (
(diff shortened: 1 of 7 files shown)


laserson:
Manually merged in c83a116a99.  Thanks @lskuff and @nonsleepr!

PR #110, opened 2015-08-21
Bug fixes for Ibis-on-Python3

wesm:
There's some problems with the null handling out of thriftpy. See ibis@refactor/python3-support branch in Ibis

cross ref: https://github.com/cloudera/ibis/pull/568

### DIFF
--- a/impala/_rpc/hiveserver2.py
+++ b/impala/_rpc/hiveserver2.py
@@ -27,10 +27,8 @@ from __future__ import absolute_import, division
 import datetime
 import socket
 import operator
-import os
 import re
 import six
-import sys
 from decimal import Decimal
 from six.moves import range
 
@@ -154,8 +152,8 @@ def connect_to_impala(host, port, timeout=45, use_ssl=False, ca_cert=None,
         sock.setTimeout(timeout * 1000.)
     elif six.PY3:
         sock.set_timeout(timeout * 1000.)
-    transport = get_transport(sock, host, kerberos_service_name, auth_mechanism,
-                              user, password)
+    transport = get_transport(sock, host, kerberos_service_name,
+                              auth_mechanism, user, password)
     transport.open()
     protocol = TBinaryProtocol(transport)
     if six.PY2:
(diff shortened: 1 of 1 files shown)


impala/_rpc/hiveserver2.py
```diff
@@ -226,10 +224,46 @@ def get_result_schema(service, operation_handle):
...
 
 
+_re_nulls = re.compile(b'^(\x00)+$')
+
+
+if six.PY3:
+
+    def _get_null(values, nulls, i):
+        if not isinstance(nulls, bytes):
+            nulls = nulls.encode('utf8')
```
wesm: BTW this is total craziness. in some cases `nulls` is bytes and in others unicode. Must be a bug in thriftpy

impala/_rpc/hiveserver2.py
```diff
@@ -226,10 +224,46 @@ def get_result_schema(service, operation_handle):
...
+        # int repr and pull out the bit at the corresponding pos
+        is_null = False
+        # Ref HUE-2722, HiveServer2 sometimes does not add not put
+        # trailing '\x00'.
+        if len(values) != len(nulls):
+            nulls = nulls + (b'\x00' * (len(values) - len(nulls)))
+        # Hive encodes nulls differently than Impala
+        # (\x00 vs \x00\x00 ...)
+        if not _re_nulls.match(nulls):
+            is_null = bool(nulls[i // 8]) & (1 << (i % 8))
```
laserson: The `bool` seems suspect to me here.  Doesn't this destroy all the info in the byte you're pulling out?

wesm: the result of `nulls[i // 8]) & (1 << (i % 8)` is an integer, I think (this is different from Python 2). It's 0 if the value is not null (

wesm:
```
>>> ord('\x00')
0
>>> ord('\x01')
1
```

laserson: I'm confused by your parentheses.

wesm: That's a typo

wesm: fixing parenthesization and the failure went away.

laserson: Is this not sufficient?

https://github.com/cloudera/impyla/pull/111

wesm:
OK, cloudera/ibis#568 is now green on top of this PR. If #111 is enough to fix it let me know

laserson:
Closed in favor of #111

Issue #114, opened 2015-08-26
Columnar HS2 row fetching is extremely inefficient


wesm:
Much work is being duplicated to handle idiosyncrasies with Hive. The nulls should be fixed once per column rather than inside the inner loop (so we are performing O(rows * cols) work rather than simply O(cols) work to fix the nulls)

PR #115, opened 2015-08-26
Improve HS2 columnar fetch performance

wesm:
per #114. Much work was being duplicated on the inner loop before.

### DIFF
--- a/impala/hiveserver2.py
+++ b/impala/hiveserver2.py
@@ -18,6 +18,7 @@ import six
 import time
 import getpass
 
+from impala.compat import lzip, Decimal
 from impala.util import get_logger_and_init_null
 from impala.interface import Connection, Cursor, _bind_parameters
 from impala.error import NotSupportedError, OperationalError, ProgrammingError
@@ -26,7 +27,6 @@ import datetime
 import socket
 import operator
 import re
-from decimal import Decimal
 from six.moves import range
 
 from impala.error import HiveServer2Error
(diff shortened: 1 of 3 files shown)


impala/hiveserver2.py
```diff
@@ -686,43 +690,56 @@ def fetch_results(service, operation_handle, hs2_protocol_version, schema=None,
     err_if_rpc_not_ok(resp)
 
     if hs2_protocol_version == TProtocolVersion.HIVE_CLI_SERVICE_PROTOCOL_V6:
+        from bitarray import bitarray
```
laserson: why is this being imported here?

impala/hiveserver2.py
```diff
@@ -686,43 +690,56 @@ def fetch_results(service, operation_handle, hs2_protocol_version, schema=None,
...
+            values = tcols[j].values
+
+            if six.PY3 and isinstance(nulls, str):
+                # thriftpy sometimes returns unicode instead of bytes
+                nulls = nulls.encode('utf-8')
+
+            # HACK: Hive hack; needs tests
+            # Hive encodes nulls differently than Impala
+            # (\x00 vs \x00\x00 ...)
+            if not re.match(b'^(\x00)+$', nulls):
```
laserson: I think we can drop the regular expression as long as the nulls string is long enough

laserson:
Merged from #115

PR #133, opened 2015-09-18
Make setup.py compatible with older versions of setuptools

jkukul:
This patch makes it possible to install `impyla` with older versions of `setuptools` (versions which didn't have `include` keyword in `find_packages` method). Installation logic remains unchanged.

Although a relatively new version of `setuptools` (`3.4.4`) is explicitly installed in the beginning of `setup.py`, on systems where `setuptools` is installed via `.deb` package, the latter will be used, since it's higher in the system hierarchy (and it can be pretty old).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -36,9 +36,7 @@ reqs = ['six', 'thrift_sasl', 'bitarray']
 packages = find_packages(exclude=['impala._thrift_gen',
                                   'impala._thrift_gen.*'])
 if PY2:
-    apache_thrift_pkgs = find_packages(include=['impala._thrift_gen',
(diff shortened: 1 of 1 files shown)


laserson:
Thanks for the patch, @jkukul!  As a formality, I would need to get a CLA from you before I can merge it.  Email me at my last name at cloudera dot com.  I'd also be fine making the change myself, if you don't want to jump through the hoops.

jkukul:
@laserson - I'm totally fine with you making the change, just go for it.

laserson:
done.  thanks for pointing this out!

PR #142, opened 2015-10-16
Call close_operation() during cursor.close()

caseyching:
If the operation isn't closed, the server will cancel the operation
which could abort a DDL op.


laserson:
Sorry for the delay, @caseyching.  Running the Jenkins tests now.

laserson:
Manually merged.  Thanks, @caseyching!

PR #145, opened 2015-10-28
MAINT: Remove deprecated functionality

wesm:


impala/dbapi.py
```diff
@@ -24,9 +24,7 @@
     Error, Warning, InterfaceError, DatabaseError, InternalError,
     OperationalError, ProgrammingError, IntegrityError, DataError,
     NotSupportedError)
-from impala.util import (
-    warn_deprecate_hs2, warn_deprecate, warn_deprecate_protocol)
-import impala.beeswax as beeswax
+from impala.util import warn_deprecate, warn_deprecate_protocol
```
laserson: Could you modify `warn_deprecate_protocol` to say that it's no longer necessary because HiveServer2 is the only protocol supported?

And maybe something like:

```python
if protocol.lower() == 'hiveserver2':
    warn_deprecate_protocol()
elif protocol is not None:
    raise NotSupportedError('...')  # or similar
service = hs2.connect(...)
```

And finally, move the `protocol=None` parameter further to the back (with the other deprecated parameters).

laserson:
Closed in 0913f6d0faa36394b0534d0a1a52aa2a33dce667 and 7aa0741fc6ef39dd217d3d81e19e1d1f6df830be.  Thanks, @wesm!

PR #146, opened 2015-10-29
ENH: add prototype columnar HS2 fetch API for Ibis

### LINKED ISSUE #144: Expose HS2 columnar fetch API
For data consumers like pandas, the fetched data is being pivoted from columnar to row-oriented and back again. This is wholly avoidable

wesm:
Fixes #144 

First efforts at handling columnar HS2 data in a more efficient way. This whole module is dying for some refactoring love, but this will do for now.

Probably a better long-term approach will be to built a direct HS2 C++ Thrift interface to the forthcoming in-memory columnar format, but this at least will enable me to round-trip data correctly from pandas to Impala for now.

### DIFF
--- a/impala/hiveserver2.py
+++ b/impala/hiveserver2.py
@@ -322,13 +322,26 @@ class HiveServer2Cursor(Cursor):
                 break
         return local_buffer
 
-    def fetchall(self):
+    def fetchall(self, columnar=False):
         # PEP 249
         log.info('Fetching all result rows')
-        try:
-            return list(self)
-        except StopIteration:
-            return []
+
+        if not columnar:
+            try:
+                return list(self)
+            except StopIteration:
+                return []
+        else:
+            batches = []
+            while True:
+                try:
+                    batch = self._fetch_batch(columnar=True)
+                    if batch.length == 0:
+                        break
+                    batches.append(batch)
+                except StopIteration:
+                    break
+            return batches
 
     def setinputsizes(self, sizes):
         # PEP 249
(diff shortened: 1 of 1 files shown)


wesm:
Full test validation requires a forthcoming Ibis PR...

impala/hiveserver2.py
```diff
@@ -322,13 +322,26 @@ def fetchmany(self, size=None):
                 break
         return local_buffer
 
-    def fetchall(self):
+    def fetchall(self, columnar=False):
```
laserson: I feel like `columnar` (or something a bit more descriptive) should probably be a field of `HiveServer2Cursor` that gets set on instantiation, no?  It seems a bit weird to me that someone could have a cursor and read columns or rows on different calls.  What do you think?

Another alternative might be to have a totally separate "internal-ish" method that does the columnar fetch, meant only for consumers that want to integrate more deeply (e.g., pandas).

laserson: `fetch_columnar()` perhaps?

impala/hiveserver2.py
```diff
@@ -342,26 +355,29 @@ def __iter__(self):
...
-            log.debug('__next__: popping row out of buffer')
-            return self._buffer.pop(0)
-        elif self._last_operation_active:
+            raise ProgrammingError("Trying to fetch results on an operation "
+                                   "with no results.")
+        if self._last_operation_active:
             # self._buffer is empty here and op is active: try to pull more
             # rows
-            log.debug('__next__: buffer empty and op is active => fetching '
+            log.debug('fetch: buffer empty and op is active => fetching '
```
laserson: should be '_fetch_batch:'

laserson: How do you know that the buffer is empty here?  If this call to `_fetch_batch` comes after there was a preceding call to `fetchone`, the buffer might have some records in it that would get skipped.

impala/hiveserver2.py
```diff
@@ -760,11 +851,42 @@ def fetch_results(service, operation_handle, hs2_protocol_version, schema=None,
...
+
+
+def fetch_results(service, operation_handle, hs2_protocol_version, schema=None,
+                  max_rows=1024, orientation=TFetchOrientation.FETCH_NEXT,
+                  columnar=False):
+    fetcher = HS2Fetch(service, operation_handle, hs2_protocol_version,
+                       schema=schema, max_rows=max_rows,
+                       columnar=columnar,
+                       orientation=orientation)
+    return fetcher.get_results()
```
laserson: Maybe I've been looking at too much Java, but this fn will return two very different types of data depending on the `columnar` setting.  Is there a better way to unify these together?

laserson:
k, feeling pretty tired.  Will continue in the morning.

impala/hiveserver2.py
wesm: Sure, that's fine with me

laserson:
Closing in favor of #150.

PR #150, opened 2015-10-30
ENH: add columnar fetch for HS2

laserson:
Replaces #146.


laserson:
@wesm could you check over this when you get a chance?  Main points:

* There is a `fetchcolumnar` method that pulls columnar data in batches
* There is a `Batch` "interface" that is returned from `fetch_results`, so the usual row-by-row API is implemented by a `.pop()` method on these objects.

Issue #151, opened 2015-10-30
Add option to leave value conversion to consumers


wesm:
`impala.hiveserver2._parse_timestamp` is very slow; if data is going to end up in a `pandas.DataFrame`, it would be better to leave all conversions to pandas API functions

PR #153, opened 2015-11-07
Refactor HS2 code to encapsulate client, session, and operation handles

wesm:
Created wrapper classes for each of the thrift handle types to simplify code and reduce the need for parameter wrangling. 

In particular, there is now:

- 1 RPC code path
- 1 logging code path

I will also address #151 here. There will be a companion Ibis PR


wesm:
ok, jenkins passes. merging

PR #156, opened 2015-11-17
Marker replacement fix

keiichishima:
1. Stop applying different marker replacement code concurrently
that will cause problems when the previous marker replacement code
generates new marker strings used in later marker replacement codes.

2. Use the string format mechanism to replace markers with values,
instead of using the find() and replace() methods.  Such code
doesn't work when a replaced value contains another marker string.

### DIFF
--- a/impala/interface.py
+++ b/impala/interface.py
@@ -176,45 +176,50 @@ class Cursor(object):
             reraise(exc_type, exc_val, exc_tb)
 
 
+def _replace_simple_markers(marker, operation, string_parameters):
+    """
+    Replaces qmark and format markers in the given operation, from the
+    string_parameters list.
+
+    Raises ProgrammingError on wrong number of parameters or bindings.
+    """
+    param_count = len(string_parameters)
+    marker_count = operation.count(marker)
+    if marker_count == 0:
+        return operation, False
+    if param_count != marker_count:
+        raise ProgrammingError("Incorrect number of bindings supplied. "
+                               "The current statement uses %d, and "
+                               "there are %d supplied." %
+                               (marker_count, param_count))
+    operation = operation.replace(marker, '{}')
+    return operation.format(*string_parameters), True
+
+
 def _replace_numeric_markers(operation, string_parameters):
     """
-    Replaces qname, format, and numeric markers in the given operation, from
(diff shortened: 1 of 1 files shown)


impala/interface.py
```diff
@@ -176,45 +176,50 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+    param_count = len(string_parameters)
+    marker_count = operation.count(marker)
+    if marker_count == 0:
+        return operation, False
+    if param_count != marker_count:
+        raise ProgrammingError("Incorrect number of bindings supplied. "
+                               "The current statement uses %d, and "
+                               "there are %d supplied." %
+                               (marker_count, param_count))
+    operation = operation.replace(marker, '{}')
```
laserson: I don't think this is compatible with Python 2.6

laserson:
Thanks @keiichishima!  I will go through this in more detail over the next few days.  Also, could you elaborate on your point #1?  What do you mean by concurrently applying marker replacement?

keiichishima:
Hello @laserson, thank you for checking my request.

For the issue #1, I was thinking the case when a replaced value have an another kind of marker string that is not processed yet.  For example, when a sql text is something like "inert into table (time_col) values (?)" and the argument is a time string such as "2015-11-26 11:11:11", the original code first replaces "?" to "2015-11-26 11:11:11" by the first call of "operation = replace_markers('?', operation, string_parameters)" in the _replace_numeric_markers() function. The resulting value of the operation variable will be "inert into table (time_col) values (2015-11-26 11:11:11)".  Since the variable includes ":1" string, this is considered as a numeric marker and replaced again at the end of the process in the _replace_numeric_markers() function.

The similar thing happens when using a "%s" marker, and replaced value includes a time string, also in the case when using a "?" marker, and replaced string include "%s" string.

impala/interface.py
keiichishima: Ah, yes, you are correct.  I was checking with 2.7 and 3.5 branches, but didn't do with 2.6 branch...

wesm:
I think this was fixed in #208 (from bug report #206) -- if there is still a problem please open a new issue. thanks!

PR #158, opened 2015-12-03
Pass async arg so that we're actually async

### LINKED ISSUE #157: execute_async doesn't seem to be async
```python
args = parser.parse_args()

conn = connect(
    host=args.host,
    port=args.port,
    user=args.user,
    auth_mechanism="PLAIN"
)
cursor = conn.cursor()
cursor.execute_async("SELECT max(somecolumn) from sometable")

print("after execute")

while cursor.is_executing():
    print(cursor.get_log())
    print(cursor.get_summary())
    time.sleep(5)
```

When I run the above, nothing after the execute call happens until the query completes.

I assumed for a while that it was human error but this is straightforward enough that I can't see anything wrong with what I'm trying to do, as it's effectively the example provided with the pull request that added this feature. Can someone confirm this is actually a bug or if I'm insane? Running on Python 3.5.0

Raynes:
This will fix #157, but as noted in the ticket the two methods I was trying to call fail regardless of whether the query is being executed async or not.

### DIFF
--- a/impala/hiveserver2.py
+++ b/impala/hiveserver2.py
@@ -218,7 +218,8 @@ class HiveServer2Cursor(Cursor):
                 self._last_operation_string = operation
 
             op = self.session.execute(self._last_operation_string,
-                                      configuration)
+                                      configuration,
+                                      async=True)
             self._last_operation = op
 
         self._execute_async(op)


laserson:
@Raynes you've successfully run async queries against Hive with this?  Does running synchronous queries against Hive also work as expected with this change?

Also, I'm happy to merge your patch, but first we'd need to get a signed CLA from you, either individual or corporate (if you're working on behalf of your company).

Individual:
https://github.com/cloudera/Impala/wiki/Individual-Contributor-License-Agreement-(ICLA)

Corporate:
https://github.com/cloudera/Impala/wiki/Corporate-Contributor-License-Agreement-(CCLA)

This is a pretty small change, so I could also make it myself if you prefer.

wesm:
@Raynes if you could send me the correct CLA to wes @ cloudera just so we're crossing our i's and dotting our t's, much appreciated. thank you!

PR #159, opened 2015-12-08
Move a bulk of logging in hiveserver2 to debug.

ishaan:
Currently, too much information is displayed at the info log level. This patch moves most
of the messages to debug to reduce noise.


laserson:
thanks @ishaan!

PR #162, opened 2015-12-14
fix keyerror when fetching complex types data

willwangcn:
When fetching the complex types data, this method will raise KeyError Exception.
```python
tcols = [_TTypeId_to_TColumnValue_getters[schema[i][1]](col)
                 for (i, col) in enumerate(resp.results.columns)]
```

### DIFF
--- a/impala/hiveserver2.py
+++ b/impala/hiveserver2.py
@@ -501,6 +501,10 @@ _TTypeId_to_TColumnValue_getters = {
     'BINARY': operator.attrgetter('binaryVal'),
     'VARCHAR': operator.attrgetter('stringVal'),
     'CHAR': operator.attrgetter('stringVal'),
+    'MAP': operator.attrgetter('stringVal'),
+    'ARRAY': operator.attrgetter('stringVal'),
+    'STRUCT': operator.attrgetter('stringVal'),
+    'UNIONTYPE': operator.attrgetter('stringVal'),
 }
 
 _pre_columnar_protocols = [


laserson:
Thanks, @willwangcn.  I've never tested fetching of nested data with impyla, and I imagine it would continue to be broken even with your fix, as there probably needs to be some kind of "parsing" of the Thrift data structures into more pythonic ones which is not implemented at all.  This could be a problem with Hive, but it is not a problem at all for working with nested data on Impala, since the Impala query syntax requires that all result sets have flat structures.

We are still deliberating on what to do specifically wrt supporting nested query results from Hive.  Is this a feature that's important to you?

willwangcn:
@laserson , thanks for your reply. I mostly use impyla as hive client and have the requirement to fetch map or other complex types data directly.

karamazi:
Hey, I'm migrating from pyhs2 to impyla and needed the same thing, along with DATE type. This change works flawlessly. Parsing complex data types would be a plus, but getting a `string` instead of `KeyError` is a huge improvement :)

For example, query on (string, date, int, map, decimal, string, bigint) table:
```
 result = hive.fetch_all("SELECT * FROM metrics.metrics LIMIT 10")
 for r in result:
     print(r)
```
```
DEBUG: Executing query: SELECT * FROM metrics.metrics LIMIT 1
('some_metric', '2016-01-06', -1, '{"all-users":"all-users"}', Decimal('14'), 'PLN', 1452682340)
```
I had to fork and change it in our gitlab, since time is of the essence, but I guess there are many folks around who could use that.
:+1:

wesm:
@laserson +1 to merge this as a stopgap until we can implement some tests

PR #175, opened 2016-02-08
Fix KeyError when fetching DATE type.

karamazi:
Analogically to https://github.com/cloudera/impyla/pull/162
It's preferable to get string representation of DATE type rather than KeyError.


wesm:
Thanks! We will need to work on test coverage for all this at some point.

PR #180, opened 2016-02-22
Wait until execution has finished when calling cursor.fetch*

julienbaley:
The current docstring for execute_async indicates "A call to `fetch*` will
block."; this was not the case, it only appeared to do so because no execution
was ever synchronous. 1253050 fixes the lack of asynchronicity, revealing the
bug.

### DIFF
--- a/impala/hiveserver2.py
+++ b/impala/hiveserver2.py
@@ -367,6 +367,7 @@ class HiveServer2Cursor(Cursor):
 
     def fetchone(self):
         # PEP 249
+        self._wait_to_finish()
         if not self.has_result_set:
             raise ProgrammingError("Tried to fetch but no results.")
         log.debug('Fetching a single row')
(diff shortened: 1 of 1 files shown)


wesm:
Thank you. I will merge soon and cut the next release.

julienbaley:
@wesm , thanks.

By the way, there's still a problem with cancel_operation (/killing queries):
it does indeed kill the query, but both the `cancel_operation` and `is_executing` (or fetch*) raise a protocol exception (something "type=4").
For now I'm catching the exception, but I think `cancel_operation` shouldn't fail and `is_executing` should rather raise some ERROR_STATE exception.

I haven't had time to investigate it further as I can work around it.

wesm:
I see. If you're able to give a test case reproduction I can look into it, or feel free to submit a patch. thanks!

PR #186, opened 2016-03-25
remove sasl and thrift_sasl from requirments

marshall245:
successfully conda built impyla and ibis-framework without sasl, thrift_sasl dependencies after this change.


wesm:
Thank yoU!

PR #191, opened 2016-04-13
Fix SQLAlchemy connector to work with Hive

joshwalters:
Removed default execution of `select version()`, this causes a failure in Hive as the `version()` function does not exist in Hive.

I have tested and verified that this change fixes SQLAlchemy connection for Hive.

### DIFF
--- a/impala/sqlalchemy.py
+++ b/impala/sqlalchemy.py
@@ -132,7 +132,6 @@ class ImpalaDialect(DefaultDialect):
         return impala.dbapi
 
     def initialize(self, connection):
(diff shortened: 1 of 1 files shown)


joshwalters:
Also, after this is fixed, would it be possible for the maintainers to do a release? This fixed version would be very useful to the [Caravel](https://github.com/airbnb/caravel) project.

gbrian:
+1

kottameister:
👍  This would really useful for my personal project as well.

wesm:
I'm sorry for the delay. Looks good, merging and will push a minor release

Issue #200, opened 2016-06-07
Operational Error - "Operation is in ERROR_STATE"


gittify:
How do I get the error details along with this error?

wesm:
This should be easy to add (we are retrieving the error message, but discarding it). Can you produce an example query that yields ERROR_STATE?

gittify:
If we try to insert into  a table that is created as follows:

INSERT query - insert into test_table partition (dt,installtime) values ('key1','nos','type1','2016-06-08 00:00:00', '2016-08-04 01:01:01');

CREATE  TABLE test_table(
      id string,
      name string,
      type string)
     PARTITIONED BY ( 
       dt string,
       installtime string)
     ROW FORMAT DELIMITED 
       FIELDS TERMINATED BY '\001' 
     STORED AS INPUTFORMAT 
       'org.apache.hadoop.hive.ql.io.RCFileInputFormat' 
     OUTPUTFORMAT 
       'org.apache.hadoop.hive.ql.io.RCFileOutputFormat';

It throws the following Error:

 Error: java.lang.RuntimeException: org.apache.hadoop.hive.ql.metadata.HiveException: Hive Runtime Error while processing row {"tmp_values_col1":"key1","tmp_values_col2":"nos","tmp_values_col3":"type1","tmp_values_col4":"2016-06-08 00:00:00","tmp_values_col5":"2016-08-04 01:01:01"}
	at org.apache.hadoop.hive.ql.exec.mr.ExecMapper.map(ExecMapper.java:179)
	at org.apache.hadoop.mapred.MapRunner.run(MapRunner.java:54)
	at org.apache.hadoop.mapred.MapTask.runOldMapper(MapTask.java:453)
	at org.apache.hadoop.mapred.MapTask.run(MapTask.java:343)
	at org.apache.hadoop.mapred.YarnChild$2.run(YarnChild.java:163)
	at java.security.AccessController.doPrivileged(Native Method)
	at javax.security.auth.Subject.doAs(Subject.java:415)
	at org.apache.hadoop.security.UserGroupInformation.doAs(UserGroupInformation.java:1671)
	at org.apache.hadoop.mapred.YarnChild.main(YarnChild.java:158)
[...]

PR #211, opened 2016-09-07
Return detailed Error Messages from query failures

schaffino:
Allow the usage of the errorMessage field from the GetOperationStatus command in an exception on impala 2.7+ where IMPALA-1633 has been implemented. This will fix issue #200.

Below commit would provide this functionality for existing impala versions but has the side effect of  invalidating the query handle and breaking subsequent calls to get_log which users may rely on for extra information after a query failure.
https://github.com/schaffino/impyla/commit/cea034545611127945114e8d57d1bc122e8db1ba

### DIFF
--- a/impala/hiveserver2.py
+++ b/impala/hiveserver2.py
@@ -312,12 +312,17 @@ class HiveServer2Cursor(Cursor):
     def _wait_to_finish(self):
         loop_start = time.time()
         while True:
-            operation_state = self._last_operation.get_status()
+            req = TGetOperationStatusReq(operationHandle=self._last_operation.handle)
+            resp = self._last_operation._rpc('GetOperationStatus', req)
+            operation_state = TOperationState._VALUES_TO_NAMES[resp.operationState]
 
             log.debug('_wait_to_finish: waited %s seconds so far',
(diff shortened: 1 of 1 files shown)


wesm:
@toddlipcon I think this solves your problem

wesm:
@schaffino it would be nice to get the error message in existing impala versions (particularly 2.3, 2.5). i don't have a sense of how disruptive https://github.com/schaffino/impyla/commit/cea034545611127945114e8d57d1bc122e8db1ba would be, though, if you have some sense

schaffino:
It would be code breaking for some users as it breaks any further operations on the failed query handle i.e get_profile, get_summary, get_log. If people have written code that uses these for additional information after a failure than that code would no longer work. For the majority of users though I think it would probably be beneficial.

An alternative that doesn't have the same effect could be to parse the the output of get_profile. It's a bit of a cludge though. Do you happen to know if the profile output stays consistent or can think of any situations where using and parsing get_profile would be a bad idea?

wesm:
I think this is an acceptable compromise to get the error log information. Merging this and we'll see how it goes

Issue #161, opened 2015-12-12
Address differences in Hive and Impala API


laserson:
Reported by @Raynes.

>if I do let the get_log and get_summary methods run, I get the following errors:

```
thrift.Thrift.TApplicationException: Invalid method name: 'GetExecSummary'
```

>and for get_log

```
thriftpy.thrift.TApplicationException: Invalid method name: 'GetLog'
```

>EDIT: I see. There's a different API in Hive for getting query execution logs.

FlorianWilhelm:
+1 This would be really nice to have!

bkyryliuk:
+1

Issue #160, opened 2015-12-08
Cannot access the query profile or error logs if it fails.


ishaan:
Use Case:

  Run a query through impyla, and the query fails. I want to get some additional information, namely:
     - The error log (via cursor.get_log())
     - The query profile (via cursor.get_profile())

Problem:
  I get the following error:

In [17]: cursor.get_log()
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-17-6b544831d9f7> in <module>()
----> 1 cursor.get_log()

/home/ishaan/cmf/systest/target/env/src/impyla/impala/hiveserver2.pyc in get_log(self)
    396 
    397     def get_log(self):
--> 398         return self._last_operation.get_log()
    399 
    400     def get_profile(self):

AttributeError: 'NoneType' object has no attribute 'get_log'

  
 Both of these methods internally call _last_operation.get_{log, profile}()
 However, when a query fails, _last_operation is set to None (which is should not be).

julienbaley:
@wesm can we get some help on this issue? This problem still exists in 0.13.7, except the _last_operation is not None anymore. When calling .get_log() on the cursor after a failed query, thriftpy sends ```thriftpy.thrift.TApplicationException: Invalid method name: 'GetLog'```

This is a problem for us as we let user write their own SQL and we'd like to let them know what went wrong with their query..

PR #212, opened 2016-09-08
Logging function get_log() works for Hive

FlorianWilhelm:
Right now if you connect with impyla to Hive and issue the ``get_log()`` method on a cursor, you get:

> thriftpy.thrift.TApplicationException: Invalid method name: 'GetLog'

This is due to the fact that Hive uses another API for the retrieval of logs as answered in this [stackoverflow question](http://stackoverflow.com/questions/32530086/how-realtime-capture-logs-of-query-from-hiveserver2-with-python-client). 
This PR solves the issue by implementing and calling the necessary Hive log API when you run ``get_log()`` and solves the issues #160 and #161 as well as the remaining bits of #200 (scroll down).

### DIFF
--- a/impala/_thrift_api.py
+++ b/impala/_thrift_api.py
@@ -37,6 +37,7 @@ if six.PY2:
     from thrift.transport.TSocket import TSocket
     from thrift.transport.TTransport import (
         TBufferedTransport, TTransportException)
+    from thrift.Thrift import TApplicationException
     from thrift.protocol.TBinaryProtocol import (
         TBinaryProtocolAccelerated as TBinaryProtocol)
 
@@ -58,7 +59,7 @@ if six.PY2:
 if six.PY3:
     # import thriftpy code
     from thriftpy import load
-    from thriftpy.thrift import TClient
+    from thriftpy.thrift import TClient, TApplicationException
     # TODO: reenable cython
     # from thriftpy.protocol import TBinaryProtocol
     from thriftpy.protocol.binary import TBinaryProtocol  # noqa
@@ -105,7 +106,7 @@ def get_socket(host, port, use_ssl, ca_cert):
                 return TSSLSocket(host, port, validate=False)
             else:
                 return TSSLSocket(host, port, validate=True, ca_certs=ca_cert)
-        else:    
+        else:
             from thriftpy.transport.sslsocket import TSSLSocket
             if ca_cert is None:
                 return TSSLSocket(host, port, validate=False)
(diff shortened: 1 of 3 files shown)


Dawiss:
Hi Florian, I was desperately searching for a solution to get the logs information from Hive and run into the same problem where the `GetLog` method was invalid. I tried the changes in Your PR, and I ended up with another problem that the `fetchType` argument is not present in the `TFetchResultsReq` or more specifically:
``
got an unexpected keyword argument 'fetchType'
``

This got me wondering, and I checkout the code Hue uses to acquire MR logs in real-time. It turns out that the `GetLog` is a legacy for old Hive versions(https://github.com/cloudera/impyla/blob/master/impala/thrift/TCLIService.thrift#L1179-L1180). So I changed the Thrift code the impyla uses to the one present in Hue, and also copyied over the method they use for retrieving logs(https://github.com/cloudera/hue/blob/808dd74395803ab1eb0ee7e84f5b92591bf9852c/apps/beeswax/src/beeswax/server/hive_server2_lib.py#L845-L854) and I'm able to get the logs. This rises a question if all the changes You've made are really necessary to make this work.

FlorianWilhelm:
@DavisJaunzems Hm... interesting, if I understand you correctly you are saying that ``GetLog`` is the legacy part and you readded it again by inserting ``TGetLogResp GetLog(1:TGetLogReq req);``. But wouldn't that mean that the ``fetchType``-trick is the new way then? 
Another thing, are you using Python 2.x? I tested it with Python 3 only and it worked. You should actually not get the error you are seeing independent of your Hive version since this happens only when my change in the thrift file is not correctly applied. Could you try to first uninstall impyla then install this PR branch with ``python setup.py install`` to make sure the changed thrift files are really applied correctly. Thanks.

DavisJaunzems:
@FlorianWilhelm I removed the impyla and re-installed it from Your PR, and for both Hive and Impala the fix works, and I'm able to get the logs in real-time. In my previous approach I missed the thrift file changes You've made. I compared the thrift file in You PR and the one they use in Hue(apart from adding `namespace py impala._thrift_gen.TCLIService`):
Hue:
``
// The type of a fetch results request. 0 represents Query output. 1 represents Log
  4: optional i16 fetchType = 0
``

impyla:
``
// Specify the fetchType for retrieval of logs use 1
  4: optional i16 fetchType
``

Hope this gets merged soon, thanks!

FlorianWilhelm:
Hi @laserson, it would be nice if you could comment on my PR. It seems some people find it really useful. Thanks for your time.

laserson:
@FlorianWilhelm, I no longer maintain this code at all.  I just lurk occasionally.  I'd ping @wesm for this.

wesm:
Seems reasonable. Merging and we'll see how it works for people

PR #216, opened 2016-09-29
Autoclose HiveServer2 connections

Saluev:
Now connections are closed on delete.

### DIFF
--- a/impala/hiveserver2.py
+++ b/impala/hiveserver2.py
@@ -57,6 +57,9 @@ class HiveServer2Connection(Connection):
         self.service = service
         self.default_db = default_db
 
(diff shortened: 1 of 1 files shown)


wesm:
Is `close()` idempotent? What if it has been explicitly closed by the user?

Saluev:
@wesm Yes, I've tested that scenario. My HiveServer2 connections were closing without errors for multiple times in a row.

wesm:
thanks!

Saluev:
By the way, I've figured out that `connection.close()` doesn't actually close anything. Cursors and queries continue to execute successfully after that.

PR #217, opened 2016-09-29
Forces thriftpy be 0.3.5

pecota:
Because impyla crashes when thriftpy version is greater than 0.3.5.
```
  File "/usr/local/lib/python3.4/dist-packages/impala/dbapi.py", line 28, in <module>
    import impala.hiveserver2 as hs2
  File "/usr/local/lib/python3.4/dist-packages/impala/hiveserver2.py", line 32, in <module>
    from impala._thrift_api import (
  File "/usr/local/lib/python3.4/dist-packages/impala/_thrift_api.py", line 60, in <module>
    from thriftpy import load
ImportError: cannot import name 'load'
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,7 +39,7 @@ if PY2:
 elif PY3:
     packages = find_packages(exclude=['impala._thrift_gen',
                                       'impala._thrift_gen.*'])
-    reqs.append('thriftpy>=0.3.5')
+    reqs.append('thriftpy==0.3.5')
 
 
 setup(


wesm:
Is there some other solution (like fixing the impyla usage of thriftpy)?

pecota:
@wesm That isn't really the problem. I forced my project to use thriftpy 0.3.5, but the problem still occurring. The problem only stoped when i removed and reinstall all the pip requirements. I'm closing this PR, sorry about that.

PR #218, opened 2016-09-30
Autoclose HiveServer2 sessions

Saluev:
Apparantly closing HiveServer2 connections doesn't solve the problem of hanging connections, since Hive session is created for every cursor. Auto-closing cursors is also necessary. This change makes `cursor.close()` operation idempotent and makes it called on delete.


wesm:
LGTM. thanks

Issue #219, opened 2016-10-05
thrift generate is requred for TFetchResultsReq


hikoz:
fetchType is not exists in current _thrift_gen.
you need to re-generate thrift files.

https://github.com/cloudera/impyla/commit/2397ef2ce4d4b07dbb358661489245fa51328d39#diff-d3d6c567895a72663f13e753e636d355R1025
https://github.com/cloudera/impyla/blob/master/impala/_thrift_gen/TCLIService/ttypes.py#L5762

```
  File "xx/site-packages/impyla-0.13.8-py2.7.egg/impala/hiveserver2.py", line 1033, in get_log
    fetchType=1)
TypeError: __init__() got an unexpected keyword argument 'fetchType'
```

schaffino:
#220 Not tested with python3 but seems to work fine with python 2.

PR #220, opened 2016-10-11
Fixes for #219 and #211 and revert #216

schaffino:
Fixes issue with hive get logs #219 
"TypeError: __init__() got an unexpected keyword argument 'fetchType' 
by using newer  TCLIService bindings from hue.

d84e8ef9ff24e395aa0105a080aad32486c46176 from #216 has been removed as this was closing the session and breaking active queries. If a connection object is being torn down that doesn't mean that we don't still have active cursors.

Fix for #211 has been added for versions of impala pre and post 2.7.0 with a flag to turn it off for users that want to retain previous functionality.


wesm:
Thank you for this. I will try to cut an impyla release in the coming weeks

Issue #221, opened 2016-10-25
Query performance 1 minute in Hue Impala vs 2 hours in python fetchall()


ragesz:
In Hue Impala my query runs less than 1 minute, but (exactly) the same query using impyla runs more than 2 hours. The python script runs on the same machine where the Impala daemon runs. My query is a simple "SELECT * FROM my_table WHERE col1 = x;" . The data is (Parquet) partitioned by "col1".

The performance of the cluster is getting high only on the first half minute after Python script starts. The size of my query result is at about 1GB but the memory usage of my Python script increases continuously from some hundred MB until at about 15GB.

In the Python code `cursor.execute(sql_query)` finishes in less than 20 seconds (`sql_query` is the above query), but `res = cursor.fetchall()` runs for ~2 hours.

Does anybody know what the problem could be?
Thank you!

schaffino:
I would imagine hue only fetches enough of your data to display and would be much slower if you tried to pull back the full result set. Python is never going to be the fastest way to pull back large amounts of data. For bulk exports of data your best bet is to write the table out as text and read the actual data from hdfs. If you want to save bandwidth you could also write out the result set compressed and pull that back. LZO is good for this as its not particularly taxing for cpu on the client side so the bottleneck is usually bandwidth when fetching the data.

There does seem to be something weird going on in impyla though and it seems to be related to type conversion. If i set  convert_types=False on the cursor, testing on a table with a million rows with a couple timestamp and decimal columns my run time goes from 90 seconds to 15. This is understandable as profiling the codes shows loads of time spent in decimal and strptime functions. However, strangely enough when this is disabled my cursor is still returning decimals and datetime objects perfectly fine.

@wesm is this taken care of upstream now so the option should be removed from impyla altogether?

Additionally, there are some other ways of speeding things up and not using so much resource.

If you don't need all the data in memory at once and you are doing some other operation that is more likely to be the bottleneck it makes sense to use the cursor as an iterator which you can loop over rather than wait for the whole result set to be pulled back.

```python
sql_query='select * from some_db.some_table'
imp_cur.execute(sql_query)
for row in imp_cur:
    do_someting    
```

There is also the fetchcolumnar method which returns batches of rows. This is about 2.5 times as fast for me but you have to pull the full resultset back as there doesn't currently seem to be a builtin way to  iterate over batches as they are received.

```python
sql_query='select * from dome_db.some_table'
imp_cur=get_cursor('impala')
imp_cur.execute(sql_query)
[...]

PR #223, opened 2016-10-28
Usability and speed improvements for fetching results

schaffino:
Relating to issue #221 

    * Enable usage of convert_types outside of fetchcolumnar function.
    * Add fetchcbatch function so that you can work on columnar data without pulling the full result set.
    * pop(0) calls were a significant cause of slowdown in the materialisng rows from the cBatch object. These have been swapped for index based access and i'm seeing up large performance increases, especially for large or wide tables.

Additionally fixed some commenting issues.


wesm:
@schaffino this introduced a failure according to a user https://github.com/cloudera/impyla/commit/0dee44c53a8eb0123123d227c86bc93546ddfa35#commitcomment-19690919 can you have a look? see #225

PR #231, opened 2016-12-11
Add partitions to table index like PyHive does

korhner:


impala/sqlalchemy.py
```diff
@@ -182,9 +182,27 @@ def get_foreign_keys(self, connection, table_name, schema=None, **kw):
...
+        # Strip whitespace
+        rows = [[col.strip() if col else None for col in row] for row in rows]
+        # Filter out empty rows and comment
+        rows = [row for row in rows if row[0]]
+        for i, (col_name, _col_type, _comment) in enumerate(rows):
+            if col_name == '# Partition Information':
+                break
+        # Handle partition columns
+        col_names = []
+        for col_name, _col_type, _comment in rows[i + 1:]:
```
timarmstrong: How does this work? It seems like this is going to process all of the "DESCRIBE FORMATTED" output after "#Partition Information" regardless of whether it's a comment or not.

korhner: My best guess is # Partition Information was last in describe output at that time in impala. Will add a fix in the next few days when I catch some time.

timarmstrong:
Thanks Ivan!

Issue #235, opened 2017-01-16
AttributeError: 'TBufferedTransport' object has no attribute 'trans'


crazy2323:
conn = connect(host='my host', port=10000, database=db, auth_mechanism='NOSASL')
cur = conn.cursor()
  File "/usr/local/python27/lib/python2.7/site-packages/impala/hiveserver2.py", line 125, in cursor
    session = self.service.open_session(user, configuration)
  File "/usr/local/python27/lib/python2.7/site-packages/impala/hiveserver2.py", line 995, in open_session
    resp = self._rpc('OpenSession', req)
  File "/usr/local/python27/lib/python2.7/site-packages/impala/hiveserver2.py", line 923, in _rpc
    response = self._execute(func_name, request)
  File "/usr/local/python27/lib/python2.7/site-packages/impala/hiveserver2.py", line 940, in _execute
    return func(request)
  File "/usr/local/python27/lib/python2.7/site-packages/impala/_thrift_gen/TCLIService/TCLIService.py", line 175, in OpenSession
    return self.recv_OpenSession()
  File "/usr/local/python27/lib/python2.7/site-packages/impala/_thrift_gen/TCLIService/TCLIService.py", line 193, in recv_OpenSession
    result.read(self._iprot)
  File "/usr/local/python27/lib/python2.7/site-packages/impala/_thrift_gen/TCLIService/TCLIService.py", line 1109, in read
    fastbinary.decode_binary(self, iprot.trans, (self.__class__, self.thrift_spec))
AttributeError: 'TBufferedTransport' object has no attribute 'trans'

is anybody help me?

czxm:
Please try below to fix this issue by enforcing thrift-0.9.3:
pip install thrift=0.9.3

https://github.com/cloudera/impyla/issues/234

crazy2323:
it works  well done thx @czxm !

mingzhou:
it works for me. thanks @czxm !

tdhopper:
Is there a reason _not_ to pin it [here](https://github.com/cloudera/impyla/blob/c0b06cfff099939c012e55418138274436278c6f/setup.py#L38) until this is fixed?

wesm:
No reason. Can someone make a PR?

tdhopper:
@wesm Yes.

qinzl1:
perfect fix my problem

PR #246, opened 2017-02-13
Pin Python thrift at <= 0.9.3

tdhopper:
[thrift 0.10.0](https://pypi.python.org/pypi/thrift/0.10.0) package has a breaking change, closes https://github.com/cloudera/impyla/issues/235.

Currently installing `impyla` will require user to manually downgrade thrift to 0.9.3. This PR prohibits thrift 0.10.0 from being installed.


wesm:
Thanks @tdhopper!

PR #248, opened 2017-02-14
add support for optional krb_host parameter in connection

manuvaldes:
If host is registered under a different name in kerberos, we need to differentiate these two names (much like in pyhs2).


bmillemathias:
Hello,

just saying the patch is working and fix authentication with kerberos when krb host is ! host.
Waiting for the patch to be commited.

Regards

timarmstrong:
This is basically analogous to the change we made in IMPALA-7130 for impala-shell.

PR #251, opened 2017-03-14
Update thrift bindings and use Apache thrift everywhere.

matthewwardrop:
Recently I have had a lot of trouble with the impyla HiveServer2 database connection (primarily in Python 3), whereby the thrift transport would for some reason hang for long periods. This was especially pronounced after you cancelled a long running query. Just in case it is relevant, I am connecting to Hive 0.13. Using thriftpy there were three failure modes:
 - Thrift transport would hang waiting for data to come through the socket.
 - Session handle would not be set properly on the response, resulting in an error.
 - Transport would fail to open three times, resulting in a failure to open sessions.

This PR is the result of me attempting to find a quick fix, and after several attempts, the greatest stability was achieved by using Apache `thrift` everywhere instead of `thriftpy`, which had several weird issues regarding session handles if connections were terminated. The connection issues have not entirely disappeared, but they seem a lot better now. Still relatively frequently it seems to hang when trying to open a thrift session. Will continue to look to improve things as I get the chance. So far, it looks like the second and third failure modes above are fixed, or happen with greatly reduced frequency. Still have issues with the first.

As part of this search for a fix, I updated the thrift bindings using thrift 0.10.0.

As it is, this PR is not ready to merge, but before I went to the effort of cleaning it up, I wanted to check whether this is something that you would be interested in. @wesm

@danfrankj You might be interested in confirming the increased stability I report here.


wesm:
Updating to Thrift 0.10.0 makes sense to me. Can you also update the SASL interface in https://github.com/cloudera/thrift_sasl to work for Thrift 0.10?

matthewwardrop:
@wesm Done.

Did you want me to clean this code up, removing the commented out code?

As an aside, I've ported `pyhive` to Python 3 and it is presently much more reliable (it never hangs). Of course, it also lacks support for various authentication schemes. I don't have a huge amount of time to dig into this, but it might be worth tracing through and seeing what is different between the two implementations at some point.

wesm:
@matthewwardrop unfortunately I'm in a position where I don't have any time to invest in Impyla, would you like to help maintain the project (same for thrift_sasl)? I believe @gregrahn or someone from Cloudera will likely be able to give you commit rights if you fill out a CLA and promise to do no harm. I mean, I can push the green buttons but I would feel guilty not putting in the energy to validate the patches against a real cluster.

matthewwardrop:
@wesm @gregrahn I too am not going to have a great deal of time to review other people's code (I'm already spread pretty thinly among lots of projects), but I have tested this code on my machine connecting to a Hive database using NOSASL locally. PyHive has a good set of unittests, and so should be able to reliably test the thrift-sasl changes. My main motivation to get impyla/pyhive working well is to make https://github.com/airbnb/omniduct reliably working in the vast majority of use-cases.

If you'd prefer to give me commit access, so that I can directly work to fix issues that come up through people's use of omniduct, I can certainly guarantee I will do no (intentional) harm... but I cannot guarantee that I will be able to devote a substantial amount of time looking at other people's submissions.

wesm:
@gregrahn if you would like to give @matthewwardrop the commit bit, then I will leave you all to it. Thanks

matthewwardrop:
@wesm Hi again! The thrift_sasl package patches are just about ready to go (just testing it against unit tests in PyHive), but in order to release it as a new package, I'd need to be about to update the package at PyPI. Is that something you want to continue doing? Or do you want to set up auto package uploading with Travis? Or do you want to add me as a maintainer on PyPI?

Also, do you have a sense of whether it is worth maintaining pre thrift 0.10.0 compatibility (which introduced Python 3 support)? I suppose it's probably a good idea.

mariusvniekerk:
I would probably advise that before this gets merged in to get rid of all the commented out py2/3 pieces.

matthewwardrop:
@mariusvniekerk For sure :).

timarmstrong:
It looks like we switched to Apache thrift >= 0.9.3 in #303 and refreshed the thrift files in 655314efda880fe20a16a8e4c01bf15fdc2a1c48, but parts of this are still relevant @lekv

lekv:
With the switch to thriftpy2 in #342 I'd be curious if the original hang still reproduces. CC @dknupp

ethe:
Hi, could you please tell me how to reproduce your issue when using thriftpy and give me more details, I will try to fix it, thanks.

csringhofer:
Another pull request implements a very similar thing (as a followup work after upgrading to Thrift 11) :
https://github.com/cloudera/impyla/pull/439

I think that this can be closed as duplicate.

matthewwardrop:
That sounds good to me. :).

PR #257, opened 2017-05-09
Add 'NULL' to _TTypeId_to_TColumnValue_getters (fixes #256)

### LINKED ISSUE #256: KeyError: 'NULL' when querying Spark ThriftServer where a result column has NULL type
I am using `impyla` to query Spark's Hive ThriftServer using a query roughly like the following:

```
SELECT MIN(NULL) OVER () AS int_col from table_1 t1
```

This results in the following stacktrace inside of `impyla`:

```
  File "/Users/joshrosen/anaconda/lib/python2.7/site-packages/impala/hiveserver2.py", line 393, in fetchmany
    local_buffer.append(next(self))
  File "/Users/joshrosen/anaconda/lib/python2.7/site-packages/impala/hiveserver2.py", line 437, in next
    return self.__next__()
  File "/Users/joshrosen/anaconda/lib/python2.7/site-packages/impala/hiveserver2.py", line 450, in __next__
    self.buffersize)
  File "/Users/joshrosen/anaconda/lib/python2.7/site-packages/impala/hiveserver2.py", line 1047, in fetch
    convert_types=convert_types)
  File "/Users/joshrosen/anaconda/lib/python2.7/site-packages/impala/hiveserver2.py", line 1052, in _wrap_results
    return CBatch(results, schema, convert_types=convert_types)
  File "/Users/joshrosen/anaconda/lib/python2.7/site-packages/impala/hiveserver2.py", line 730, in __init__
    for (i, col) in enumerate(trowset.columns)]
KeyError: 'NULL'
```

I think the right fix here is to add an entry for `NULL` to _TTypeId_to_TColumnValue_getters. The Thrift `NULL_TYPE` id is already handled elsewhere in this codebase in `_thrift_gen` `ttypes.py` in the `_VALUES_TO_NAMES` map.

This is similar to #182 but is probably easier to fix.

JoshRosen:
This patch fixes #256, an issue causing `KeyError` when a result column has `NULL_TYPE`.

Based on https://github.com/cloudera/impyla/blob/94a8eff9cda0cdb16b180c7079961449c8385997/impala/thrift/TCLIService.thrift#L375, it looks like Hive's `NULL_TYPE` gets mapped to TStringColumn, so I added an entry to fetch `stringVal` attribute from the column. I believe this is the rationale for a similar change in #162 to avoid `KeyError` for complex types (which also map to `stringVal`).

### DIFF
--- a/impala/hiveserver2.py
+++ b/impala/hiveserver2.py
@@ -685,6 +685,7 @@ _TTypeId_to_TColumnValue_getters = {
     'ARRAY': operator.attrgetter('stringVal'),
     'STRUCT': operator.attrgetter('stringVal'),
     'UNIONTYPE': operator.attrgetter('stringVal'),
+    'NULL': operator.attrgetter('stringVal'),
     'DATE': operator.attrgetter('stringVal')
 }
 


JoshRosen:
@gregrahn @dknupp, sorry to ping, but could you take a look at this old PR from last year? I stumbled across this while going through my list of open pull requests and would like to see if there's a way to get this small fix merged to clear it out of my PR queue. Thanks!

Issue #259, opened 2017-05-17
SQLAlchemy Table doesn't support storing as Kudu


dimaspivak:
The `CREATE TABLE ... STORE AS KUDU` construct needed to create a Kudu internal table from Impala can be explicitly executed from Impyla, but it would be convenient if it could be added as a dialect-specific keyword argument in the SQLAlchemy connector.

dimaspivak:
Would the project be receptive if I put this together and created a PR?

dimaspivak:
For anyone following along, this would expose a new `kwarg` when creating a `Table` instance as follows:
```
>>> import sqlalchemy
>>> engine = sqlalchemy.create_engine('impala://node-1.cluster')
>>> table = sqlalchemy.Table('sweet_new_table', sqlalchemy.MetaData(), sqlalchemy.Column('id', sqlalchemy.Integer), impala_store_as='KUDU')
>>> print(str(sqlalchemy.schema.CreateTable(table, bind=engine)))

CREATE TABLE sweet_new_table (
	id INTEGER
)
STORE AS KUDU
```

zslim:
Hi! Is this feature available when using declarative mapping?

PR #260, opened 2017-05-17
SQLAlchemy Table doesn't support storing as Kudu (#259)

dimaspivak:
* Add ImpalaDDLCompiler with post_create_table method.

* Update test_sqlalchemy.py to include new dialect-specific options.


dimaspivak:
Newest commit adds support for `TBLPROPERTIES`, which we found we needed to interact with tables created outside of Impala.

Issue #268, opened 2017-10-04
'TSocket' object has no attribute 'isOpen'


tdhopper:
I'm seeing this error for the first time:

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-5-b8033d8b7753> in <module>()
      1 impala.dbapi.connect(host=HOST, port=PORT, auth_mechanism='PLAIN',
----> 2                     user=USER, password=PASSWORD)

~/miniconda2/envs/env/lib/python3.6/site-packages/impala/dbapi.py in connect(host, port, database, timeout, use_ssl, ca_cert, auth_mechanism, user, password, kerberos_service_name, use_ldap, ldap_user, ldap_password, use_kerberos, protocol)
    145                           ca_cert=ca_cert, user=user, password=password,
    146                           kerberos_service_name=kerberos_service_name,
--> 147                           auth_mechanism=auth_mechanism)
    148     return hs2.HiveServer2Connection(service, default_db=database)
    149 

~/miniconda2/envs/env/lib/python3.6/site-packages/impala/hiveserver2.py in connect(host, port, timeout, use_ssl, ca_cert, user, password, kerberos_service_name, auth_mechanism)
    756     transport = get_transport(sock, host, kerberos_service_name,
    757                               auth_mechanism, user, password)
--> 758     transport.open()
    759     protocol = TBinaryProtocol(transport)
    760     if six.PY2:

~/miniconda2/envs/env/lib/python3.6/site-packages/thrift_sasl/__init__.py in open(self)
     65 
     66   def open(self):
---> 67     if not self._trans.isOpen():
     68       self._trans.open()
     69 

AttributeError: 'TSocket' object has no attribute 'isOpen'
```

These are versions I have installed:

```
thrift==0.9.3
thrift-sasl==0.3.0
thriftpy==0.3.9
impyla==0.14.0
```

wesm:
I just pushed thrift_sasl 0.3.0 yesterday so it must be a version incompatibility. Can you upgrade to thrift 0.10.0? We probably need a version bump in impyla to pin things properly

tdhopper:
@wesm Thanks. I actually get the same error with 0.10.0.

wesm:
OK, will the fix is to pin thrift_sasl at 0.2.1 until we get impyla working on 0.10.0 + thrift_sasl 0.3.0. I am unable to work on it today unfortunately =/

tdhopper:
Yup, that works.

[...]

PR #269, opened 2017-10-09
Pinned thrift_sasl to 0.2.1 until #268 is resolved

dreyco676:
thrift_sasl was updated and now throws an error:

`AttributeError: 'TSSLSocket' object has no attribute 'isOpen'`

Pinning is a short term fix until the issue #268 is formally resolved.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ Required:
 For Hive and/or Kerberos support:
 
 ```
-pip install thrift_sasl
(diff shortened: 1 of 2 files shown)


wesm:
Thanks, we will need push a point release. I don't have any time to maintain this project or thrift_sasl, so I would be appreciative if someone would make appropriate fixes and submit patches.

PR #273, opened 2017-10-25
include thrift_sasl in setup.py

r-richmond:
It appears that setup.py misses this dependency and if i try and use this library without running `pip install thrift_sasl==0.2.1` it errors out with

`ModuleNotFoundError: No module named 'thrift_sasl'`

I've also pinned the version pending the resolution of #268


timarmstrong:
It looks like we've documented this as an optional dependency. I don't know if there's a reason why we don't require it by default. Maybe someone with more context knows.

r-richmond:
I've updated the PR to set it up as an optional install. Should be installable now with

`pip install 'impyla[kerberos]'`

timarmstrong:
Thanks for the contribution!

Issue #262, opened 2017-05-18
Kerberos Auth connection attempt results in Server Not Found in Kerberos database error


wjdecorte:
We are trying to connect to a kerberized cluster using Impyla.  When trying to connect, we get a vague error message.

```
conn = connect(host='hiveserver2.host.com',port=10000,auth_mechanism='GSSAPI',kerberos_service_name='hive/hiveserver2.host.com@REALM.COM')

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/venv/COUNTACH/lib/python2.7/site-packages/impala/dbapi.py", line 147, in connect
    auth_mechanism=auth_mechanism)
  File "/opt/venv/COUNTACH/lib/python2.7/site-packages/impala/hiveserver2.py", line 758, in connect
    transport.open()
  File "/opt/venv/COUNTACH/lib/python2.7/site-packages/thrift_sasl/__init__.py", line 72, in open
    message=("Could not start SASL: %s" % self.sasl.getError()))
thrift.transport.TTransport.TTransportException: Could not start SASL: Error in sasl_client_start (-1) SASL(-1): generic failure: GSSAPI Error: Unspecified GSS failure.  Minor code may provide more information (Server not found in Kerberos database)
```

We are able to use the same kerberos service principle and host for Beeline and connect successfully.  I saw another post mentioning this same issue, but no resolution was given (Issue #149).

Environment:
```
CDH-5.8.3
impyla      0.14.0
sasl        0.2.1
six         1.10.0
thrift      0.10.0
thrift-sasl 0.2.1
```

Any help or suggestions would be greatly appreciated.  We are at a loss as to whether it is a bug or a configuration issue or something else.

Thanks,
Jason

wfowlks:
I am also having this issue. I was able to get it to work not using kerberos. 

After kiniting and verifying using klist:
```
Ticket cache: FILE:/tmp/krb5cc_1000
Default principal: impala@HOST.COM

Valid starting       Expires              Service principal
09/27/2017 11:34:57  09/28/2017 11:34:52  krbtgt/HOST.COM@HOST.COM
	renew until 10/04/2017 11:34:52
```

```
(venv) [centos@test-haproxy supersets]$ python sql2test.py 
Traceback (most recent call last):
  File "sql2test.py", line 15, in <module>
    ca_cert='/home/centos/supersets/TEST.pem')
[...]

PR #274, opened 2017-10-30
fix error while starting SASL client

kingbase:
While connecting, error like this was raised:

TTransportException: TTransportException(type=1, message="Could not start SASL: b'Error in sasl_client_start (-1) SASL(-1): generic failure: GSSAPI Error: Unspecified GSS failure. Minor code may provide more information (Server krbtgt/XX@XXYYZZ.HADOOP not found in Kerberos database)'")

This was caused by a sasl_client parameter setting error. And the fix works well for me.

The original issue was like this: #262

### DIFF
--- a/impala/_thrift_api.py
+++ b/impala/_thrift_api.py
@@ -152,7 +152,7 @@ def get_transport(socket, host, kerberos_service_name, auth_mechanism='NOSASL',
 
         def sasl_factory():
             sasl_client = sasl.Client()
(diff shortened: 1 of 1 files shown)


timarmstrong:
This definitely isn't the right way to fix this. It looks like you can pass in the user via the user argument of connect().

kingbase:
@timarmstrong you are right, the fix was only working in my own company, I'll close this issue.

PR #309, opened 2018-07-26
Change async to run_async for Python 3.7

alexboily-ls:
`async` is now a [reserved keyword in Python 3.7](https://docs.python.org/3/whatsnew/3.7.html#changes-in-python-behavior), causing syntax errors in `hiveserver2.py`.

The syntax errors are restricted to argument names, which were changed from `async` to `run_async`. Camel-cased `runAsync` on line 1032 was kept to avoid issues with Thrift generated code.

### DIFF
--- a/impala/hiveserver2.py
+++ b/impala/hiveserver2.py
@@ -342,7 +342,7 @@ class HiveServer2Cursor(Cursor):
 
             op = self.session.execute(self._last_operation_string,
                                       configuration,
-                                      async=True)
+                                      run_async=True)
             self._last_operation = op
 
         self._execute_async(op)
@@ -1024,11 +1024,11 @@ class HS2Session(ThriftRPC):
         req = TCloseSessionReq(sessionHandle=self.handle)
         self._rpc('CloseSession', req)
 
-    def execute(self, statement, configuration=None, async=False):
+    def execute(self, statement, configuration=None, run_async=False):
         req = TExecuteStatementReq(sessionHandle=self.handle,
                                    statement=statement,
                                    confOverlay=configuration,
-                                   runAsync=async)
+                                   runAsync=run_async)
         return self._operation('ExecuteStatement', req)
 
     def get_databases(self, schema='.*'):
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,4 @@
-bitarray==0.8.3
 ply==3.11
 six==1.11.0
 thrift==0.11.0
-thriftpy==0.3.9
+thriftpy2==0.3.11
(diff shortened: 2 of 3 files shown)


cpcloud:
Can someone merge this patch? This is blocking downstream support of Python 3.7

requirements.txt
```diff
@@ -0,0 +1,4 @@
+ply==3.11
```
cpcloud: Pinning exact versions has the potential to cause a lot of headaches for downstream users of this library, especially across major Python versions.

Why did you add a requirements file?

impala/hiveserver2.py
```diff
@@ -1024,11 +1024,11 @@ def close(self):
         req = TCloseSessionReq(sessionHandle=self.handle)
         self._rpc('CloseSession', req)
 
-    def execute(self, statement, configuration=None, async=False):
+    def execute(self, statement, configuration=None, run_async=False):
```
cpcloud: If you really want to be thorough here you can maintain backwards compatibility by going through a deprecation cycle.

You would do this by setting the default value of `run_async` to `None`, adding `**kwargs` to the method signature, checking if `run_async` is not `None` and using that if so else use `kwargs.get('async', False)`. If you do decide to do this then you should also check if `'async' in kwargs` and give a warning to users that their code needs to be updated to use `run_async` because in the next release you will not accept this keyword argument.

wesm:
I can help with this -- @cpcloud are you a maintainer on this package?

cpcloud:
@wesm I am not

wesm:
@gregrahn can someone from Cloudera add @cpcloud as an admin on this project? It isn't being maintained very much, but we can at least merge incoming PRs and cut releases to PyPI

gregrahn:
Added @cpcloud -- ping me with any issues.

cpcloud:
@wesm @gregrahn Thanks guys. I'm cranking out a fix for the Python 3.7 issue. Can you give me access to PyPI as well? My username is Charles.Cloud

impala/hiveserver2.py
cpcloud: @wesm Actually, are you aware of whether this method `HS2Session.execute` is part of the public API? It looks like it might not be in which case we don't have to handle backwards compatibility here.

cpcloud: After digging around a bit I don't think this method is part of the public API. Users go through `connect -> HiveServer2Connection -> .cursor() -> HiveServer2Cursor -> execute/execute_async` and never see the `HS2Session.execute` calls.

cpcloud:
Closing in favor of #322

cpcloud:
@alexboily-ls I preserved your commits in #322.

wesm:
I think it's OK to ignore backwards compatibility issues here

PR #314, opened 2018-09-18
[Bug-fix] fix VARCHAR, DATE mismatch in hive side

yiakwy:
A dirty quick fix to datatype problems of Hive 2.x backends.

### DIFF
--- a/impala/sqlalchemy.py
+++ b/impala/sqlalchemy.py
@@ -176,11 +176,28 @@ class ImpalaDialect(DefaultDialect):
         cursor.fetchall()
         column_info = []
         for col in schema:
-            column_info.append({
-                'name': col[0],
-                'type': _impala_type_to_sqlalchemy_type[col[1]],
-                'nullable': True,
-                'autoincrement': False})
+            try:
+                column_info.append({
+                    'name': col[0],
+                    'type': _impala_type_to_sqlalchemy_type[col[1]],
+                    'nullable': True,
+                    'autoincrement': False})
+            except KeyError as e:
+                if col[1] == 'VARCHAR':
+                    column_info.append({
+                        'name': col[0],
+                        'type': _impala_type_to_sqlalchemy_type['STRING'],
+                        'nullable': True,
+                        'autoincrement': False})
+                elif col[1] == 'DATE':
+                    col[1] = 'TIMESTAMP'
+                    column_info.append({
+                        'name': col[0],
+                        'type': _impala_type_to_sqlalchemy_type['TIMESTAMP'],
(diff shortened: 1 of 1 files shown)


impala/sqlalchemy.py
```diff
@@ -176,11 +176,28 @@ def get_columns(self, connection, table_name, schema=None, **kwargs):
...
-                'nullable': True,
-                'autoincrement': False})
+            try:
+                column_info.append({
+                    'name': col[0],
+                    'type': _impala_type_to_sqlalchemy_type[col[1]],
+                    'nullable': True,
+                    'autoincrement': False})
+            except KeyError as e:
+                if col[1] == 'VARCHAR':
```
timarmstrong: Can't you add these to _impala_type_to_sqlalchemy_type?

Impala already supports VARCHAR and DATE support is upcoming so this will be useful.

impala/sqlalchemy.py
```diff
@@ -176,11 +176,28 @@ def get_columns(self, connection, table_name, schema=None, **kwargs):
...
+                    column_info.append({
+                        'name': col[0],
+                        'type': _impala_type_to_sqlalchemy_type['STRING'],
+                        'nullable': True,
+                        'autoincrement': False})
+                elif col[1] == 'DATE':
+                    col[1] = 'TIMESTAMP'
+                    column_info.append({
+                        'name': col[0],
+                        'type': _impala_type_to_sqlalchemy_type['TIMESTAMP'],
```
timarmstrong: I think this should map to a Sqlalchemy date type.

timarmstrong:
Can you add some tests for this change so we can be sure it doesn't regress in future.

yiakwy:
@timarmstrong Roger, I will refine the codes and do it soon.

impala/sqlalchemy.py
yiakwy: Totally agree.

joosbuijsNL:
Any update on this? 

In Impyla 0.16.2 we get this error (VARCHAR is not in the _impala_type_to_sqlalchemy_type list) which causes our df.to_sql call to break, with no easy/known workarounds.

joosbuijsNL:
For those with a similar issue, this on-the-fly hack seems to work:

```
import impala.sqlalchemy as impchemy
impchemy._impala_type_to_sqlalchemy_type['VARCHAR'] = sqlalchemy.types.VARCHAR

```

timarmstrong:
Yeah we need to get a proper fix for this. Will abandon and reopen a proper one.

Issue #326, opened 2019-01-25
ResourceWarning and potential socket leak in HiveServer2Cursor


ProfessorEugene:
## Problem
`HiveServer2Cursor`'s' `__del__` method / finalizer can open a new socket if the cursor has already been closed prior to garbage collection / the `__del__` method being called.

This may cause a socket leak and triggers a `ResourceWarning`s in python3.

## Explanation

Calling [`HiveServer2Cursor.__del__`](https://github.com/cloudera/impyla/blob/1ef84dba5a37c9aa4020f05895e371605513ea62/impala/hiveserver2.py#L170) can invoke the following call chain:

```
--HiveServer2Cursor.__del__
  \--H2Session.close
      \--ThriftRPC._rpc
         \--ThriftRPC._execute
            \--ThriftRPC.open_transport
               ...
               \--socket.socket(...)
```

If the socket for the underlying transport object has already been closed due to someone invoking `HiveServer2Corsor.__exit__` or `HiveServer2Corsor.close`, a new socket is created and "orphaned".  In more detail:

`HiveServer2Cursor`'s' [`__del__`](https://github.com/cloudera/impyla/blob/1ef84dba5a37c9aa4020f05895e371605513ea62/impala/hiveserver2.py#L170) method looks like this:
```python
    def __del__(self):
        try:
           self.close_operation()
        except Exception:
            pass
        try:
           self.session.close()
        except Exception:
            pass
```

The `self.session.close()` call invokes `HS2Session`'s [`close`](https://github.com/cloudera/impyla/blob/1ef84dba5a37c9aa4020f05895e371605513ea62/impala/hiveserver2.py#L1024) method, which looks like this:
```python
    def close(self):
        req = TCloseSessionReq(sessionHandle=self.handle)
        self._rpc('CloseSession', req)
```

This in turn calls `self._rpc` / [`ThriftRPC._rpc`](https://github.com/cloudera/impyla/blob/1ef84dba5a37c9aa4020f05895e371605513ea62/impala/hiveserver2.py#L927), which looks like this:
```python
    def _rpc(self, func_name, request):
        self._log_request(func_name, request)
        response = self._execute(func_name, request)
        self._log_response(func_name, response)
        err_if_rpc_not_ok(response)
        return response
```

[...]

PR #327, opened 2019-01-25
issue-326 - potential fix for a socket leak in HiveServer2Cursor

ProfessorEugene:
A potential fix for #326 along with a test that demonstrates the problem.  

`SocketTracker`, the class that keeps track of opened/closed sockets for the duration of the test seems hacky, but I can't think of a better way to do this.

### DIFF
--- a/impala/hiveserver2.py
+++ b/impala/hiveserver2.py
@@ -168,7 +168,7 @@ class HiveServer2Cursor(Cursor):
         self._closed = False
 
     def __del__(self):
-        if self._closed == True:
+        if self._closed:
             return
         try:
            self.close_operation()
--- a/impala/hiveserver2.py
+++ b/impala/hiveserver2.py
@@ -168,6 +168,8 @@ class HiveServer2Cursor(Cursor):
         self._closed = False
 
     def __del__(self):
+        if self._closed == True:
+            return
         try:
            self.close_operation()
         except Exception:
--- a/impala/tests/test_dbapi_connect.py
+++ b/impala/tests/test_dbapi_connect.py
@@ -26,7 +26,7 @@ import pytest
 
 from impala.dbapi import connect
 from impala.util import _random_id
-from impala.tests.util import ImpylaTestEnv
+from impala.tests.util import ImpylaTestEnv, SocketTracker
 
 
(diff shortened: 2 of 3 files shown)


impala/hiveserver2.py
```diff
@@ -168,6 +168,8 @@ def __init__(self, session, convert_types=True, fetch_error=True):
         self._closed = False
 
     def __del__(self):
+        if self._closed == True:
```
timarmstrong: nit: drop the == True, not needed

impala/tests/test_dbapi_connect.py
```diff
@@ -85,3 +85,22 @@ def test_impala_plain_connect(self):
     def test_hive_nosasl_connect(self):
         self.connection = connect(ENV.host, ENV.hive_port, timeout=5)
         self._execute_queries(self.connection)
+
+class ImpalaSocketTests(unittest.TestCase):
+
```
timarmstrong: Can you remove the trailing whitespace here?

impala/tests/test_dbapi_connect.py
```diff
@@ -85,3 +85,22 @@ def test_impala_plain_connect(self):
...
         self._execute_queries(self.connection)
+
+class ImpalaSocketTests(unittest.TestCase):
+    
+    def run_a_query(self):
+        with connect(ENV.host, ENV.port) as connection:
+            with connection.cursor() as cursor:
+                cursor.execute('select 1 as a_number')
+                return cursor.fetchall()
+
```
timarmstrong: Can you remove the trailing whitespace here?

impala/hiveserver2.py
ProfessorEugene: Dropped `== True` in 16decdd!

impala/tests/test_dbapi_connect.py
ProfessorEugene: Removed in 99f1d83!

impala/tests/test_dbapi_connect.py
ProfessorEugene: Removed in 99f1d83!

timarmstrong:
Thanks for your contribution Eugene!

PR #333, opened 2019-02-02
Add thrift profile support, update Impala thrift files

### LINKED ISSUE #332: Add support to retrieve query profile in different formats
The change for [IMPALA-6924](https://issues.apache.org/jira/browse/IMPALA-6924) extended the TGetRuntimeProfileReq RPC to return profiles in various formats including as structured Thrift objects. We should bump the Impala dependencies to expose that change through Impyla.

lekv:
This change exposes the new 'format' parameter in TGetRuntimeProfileReq
to allow retrieval of query profiles in different formats. It also
updates various Thrift files to more recent versions.

Testing: I used this change in a companion change to Impala that will
switch to retrieving Thrift profiles through Impyla instead of scraping
them from the debug webpages.

Closes #332


lekv:
@timarmstrong @twmarshall - Does one of you have time to review this?

lekv:
You can install this PR locally by running `pip install "git+https://github.com/cloudera/impyla@refs/pull/333/merge"`.

Before that, you might have to run `pip install --upgrade pip` to make sure your version of pip is new enough.

timarmstrong:
I can take a look

impala/thrift/process_thrift.sh
```diff
@@ -13,23 +13,29 @@
...
+}
 
-if [ -z "$IMPYLA_REPO" ]; then
-    echo "Need to set IMPYLA_REPO"
-    exit 1
-fi
+[ -n "$IMPALA_REPO" ] || die "Need to set IMPALA_REPO"
+[ -n "$IMPYLA_REPO" ] || die "Need to set IMPYLA_REPO"
+
+source $IMPALA_REPO/bin/impala-config.sh
```
timarmstrong: Can you add set -euo pipefail after source impala-config.sh . I was trying to run this and ran into issues when it kept running after failing to copy some files

impala/thrift/process_thrift.sh
```diff
@@ -39,7 +45,7 @@ grep -v 'namespace py beeswaxd' $IMPALA_REPO/common/thrift/beeswax.thrift \
 
 # hive_metastore.thrift assumes a directory structure for fb303.thrift, so we
 # change the include statement here
-cat $IMPALA_REPO/thirdparty/hive-*/src/metastore/if/hive_metastore.thrift \
+cat $IMPALA_TOOLCHAIN/cdh_components/hive-$IMPALA_HIVE_VERSION/src/metastore/if/hive_metastore.thrift \
```
timarmstrong: This doesn't work on a clean environment - cdh_components now frequently has a version after it.


$HIVE_SRC_DIR/metastore/if/hive_metastore.thrift should work

timarmstrong: After making that change there was some additional generated thrift code for the GetPartitions() method

lekv:
Thanks for the review, please see the update pull request.

impala/hiveserver2.py
```diff
@@ -964,10 +972,10 @@ def _operation(self, kind, request):
         return self._get_operation(resp.operationHandle)
 
     def _log_request(self, kind, request):
-        log.debug('%s: req=%s', kind, request)
+        log.info('%s: req=%s', kind, request)
```
timarmstrong: Is this leftover debugging code? This is going to generate a lot of noise right?

impala/hiveserver2.py
```diff
@@ -259,11 +259,13 @@ def close(self):
...
-    def cancel_operation(self):
+    def cancel_operation(self, reset_state=True):
         if self._last_operation_active:
             log.info('Canceling active operation')
-            self._last_operation.cancel()
-            self._reset_state()
+            resp = self._last_operation.cancel()
+            if reset_state:
+                self._reset_state()
+            return resp
```
timarmstrong: Do we need to expose this? Seems like we don't want clients to start to depend on this. Based on your other change this might be leftover debugging code...

impala/hiveserver2.py
```diff
@@ -394,6 +396,8 @@ def _wait_to_finish(self):
             time.sleep(self._get_sleep_interval(loop_start))
 
     def status(self):
+        if self._last_operation is None:
+            raise ProgrammingError("Operation state is not available")
```
timarmstrong: This error handling is a nice improvement.

impala/hiveserver2.py
lekv: Yes, reverted it.

impala/hiveserver2.py
lekv: I removed the return code but left the return statements inside Operation() below for consistency. Let me know if you'd prefer those removed, too.

timarmstrong:
Thank you!

Issue #281, opened 2017-11-21
Impersonation in kerberos session


venkattaku:
I'm interfacing a Hive cluster that is secured in Kerberos.
I'm able to query the cluster with kerberos authentication mechanism.
Now I want to impersonate a proxy user for my queries using impyla.
Gone through source code and realized there is no support for impersonation.

puzon0:
To impersonate another user one can pass `configuration` parameter to the `HiveServer2Connection.cursor()` method:

```python
from impala.dbapi import connect

impersonation_configuration = {
	'hive.server2.proxy.user': 'impersonated_user',  # for HiveServer2
	'impala.doas.user': 'impersonated_user'  # for Impala
}

conn = connect(
    host='hive.server.host.fqdn',
    port=server_port,
    auth_mechanism='GSSAPI',
    kerberos_service_name='hive',
    database='db_name'
)
cursor = conn.cursor(configuration=impersonation_configuration)
cursor.execute('query to execute')
```

The user connecting to the cluster (one that is authenticated using Kerberos) must be allowed to impersonate other users in the cluster configuration.

csringhofer:
Closing this as there is a workaround and there is another issue about impersonation: #456

PR #334, opened 2019-02-04
Passing username for impersonation

gbrian:


timarmstrong:
Can you add a test for this feature? Maybe you can use the effective_user() builtin function to make sure that it had the intended effect.

gbrian:
Do you mean kind of:

```
def test_sqlalchemy_impersonation():
    engine = create_engine('impala://localhost')
    result = connection.execute("select effective_user() as effective_user")
    observed = result[0]['effective_user']
    expected = '?????'
    assert expected == observed

    impersonate = 'user1'
    engine = create_engine('impala://%s@localhost' % impersonate)
    result = connection.execute("select effective_user() as effective_user")
    observed = result[0]['effective_user']
    expected = impersonate
    assert expected == observed
```

timarmstrong:
Something along those lines, yeah. I think the second part of that test is the most important. Mainly to prove that the 'effective_user' value gets plumbed all the way through to Impala correctly.

huw0:
cloudera/impyla#281 indicates that impersonation works on a per-cursor basis.

Therefore can this PR be closed? I think it will be confusing to support impersonation in multiple places and this could lead to security issues where connections are re-used.

gbrian:
Not needed

PR #337, opened 2019-02-26
Update Impala thrift files

lekv:
IMPALA-8234 fixed the order of the Metrics.TUnit enum to unbreak older
readers of Impala profiles. This commit adopts that change into Impyla.

It also makes small adjustments to process_thrift.sh to improve it's
usability.

Testing: I rebased this companion change and ran the relevant tests.
https://gerrit.cloudera.org/#/c/12530/


lekv:
@timarmstrong Do you have time to look at this?

PR #346, opened 2019-04-06
Update hiveserver2.py with new property @rowcounts

prasanna-sk:
This change adds a new property (`@rowcounts`) to cursor object. It returns modified/impacted rows and error rows.

`modifiedRows` is relevant for both kudu and non-kudu tables. 
`errorRows` is only relevant for Kudu tables. For non-kudu tables this is defaulted to -1.

### DIFF
--- a/impala/hiveserver2.py
+++ b/impala/hiveserver2.py
@@ -195,6 +195,26 @@ class HiveServer2Cursor(Cursor):
         return self._rowcount
 
     @property
+    def rowcounts(self):
+        # Work around to get the number of rows modified for Inserts/Update/Delte statements
+        modifiedRows, errorRows = -1, -1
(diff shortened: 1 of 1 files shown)


timarmstrong:
Sorry to ignore this. I looked at it and intended to come back to it but never did.

I don't particularly like the text profile parsing - it's pretty hacky, but I also don't see a way to get these values that would work with older Impala versions. So I think we can probably live with it for now.

prasanna-sk:
Agree. Wish we had a better solution.

ssubhasis:
When i call cursor.rowcounts i get error as thriftpy2.thrift.TApplicationException: Invalid method name: 'GetRuntimeProfile'. cursor.rowcount returns -1 for inserts

PR #342, opened 2019-03-06
Selectively install thriftpy2 requirement based on python version

dknupp:
This patch addresses both #329 and #326.

Tested by setting up two different virtualenvs, and confimring that
the expected requirements were installed, and that the connect function
could be imported from impala.dbapi.

**Python 2**
```
[dknupp@mbp ~/Documents/code]$ virtualenv /tmp/python2_venv
New python executable in /tmp/python2_venv/bin/python
Installing setuptools, pip, wheel...
done.
[dknupp@mbp ~/Documents/code]$ source /tmp/python2_venv/bin/activate
(python2_venv) [dknupp@mbp ~/Documents/code]$ python --version
Python 2.7.10
(python2_venv) [dknupp@mbp ~/Documents/code]$ pip list
DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date. A future version of pip will drop support for Python 2.7.
Package    Version
---------- -------
pip        19.0.3
setuptools 40.8.0
wheel      0.33.1
(python2_venv) [dknupp@mbp ~/Documents/code]$ pip install -e impyla/
DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date. A future version of pip will drop support for Python 2.7.
Obtaining file:///Users/dknupp/Documents/code/impyla
Collecting six (from impyla==0.14.2.1+8.gc010d59)
  Using cached https://files.pythonhosted.org/packages/73/fb/00a976f728d0d1fecfe898238ce23f502a721c0ac0ecfedb80e0d88c64e9/six-1.12.0-py2.py3-none-any.whl
Collecting bitarray (from impyla==0.14.2.1+8.gc010d59)
Collecting thrift>=0.9.3 (from impyla==0.14.2.1+8.gc010d59)
Installing collected packages: six, bitarray, thrift, impyla
  Running setup.py develop for impyla
Successfully installed bitarray-0.8.3 impyla six-1.12.0 thrift-0.11.0
(python2_venv) [dknupp@mbp ~/Documents/code]$ pip list
DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date. A future version of pip will drop support for Python 2.7.
Package    Version             Location
---------- ------------------- -----------------------------------
bitarray   0.8.3
[...]

PR #353, opened 2019-06-10
Make thriftpy2 range locked, not specific version locked.

vEpiphyte:
Allows for happybase and impyla to co-exist in a 3.7 environment.

Update based on @ethe note on the #342 PR


vEpiphyte:
@timarmstrong @lekv Is it possible that this can be merged and a new impyla release cut on pypi?

timarmstrong:
Thanks for the contribution. @dknupp @lekv and I have had our hands a bit full recently, so things are delayed around here. We'll look at a release.

PR #355, opened 2019-07-03
Remove non exist args from docstring

cympfh:


timarmstrong:
Thanks for the contribution!

Issue #357, opened 2019-07-22
Support HTTP endpoint in Impyla


timarmstrong:
We should extend impyla to support THttpClient so it can connect to Impala's new HTTP interface (which will be present in Impala 3.3). 

Here's an example of the code change for impala-shell https://gerrit.cloudera.org/#/c/13746/ - https://issues.apache.org/jira/browse/IMPALA-8717

lekv:
Fixed in #359

PR #359, opened 2019-08-09
Add HTTP endpoint support

lekv:
This is a preview of adding HTTP endpoint support (#357). I tested this locally against an Impala cluster on my development machine and added a trivial test to it. Additional ideas on how to test this would be very much welcome.

### DIFF
--- a/impala/_thrift_api.py
+++ b/impala/_thrift_api.py
@@ -20,10 +20,12 @@
 
 from __future__ import absolute_import
 
-import os
-import sys
-import six
+import base64
 import getpass
+import os
+import six
+import ssl
+import sys
 
 from impala.util import get_logger_and_init_null
 
@@ -34,6 +36,7 @@ log = get_logger_and_init_null(__name__)
 if six.PY2:
     # pylint: disable=import-error,unused-import
     # import Apache Thrift code
+    from thrift.transport.THttpClient import THttpClient
     from thrift.transport.TSocket import TSocket
     from thrift.transport.TTransport import (
         TBufferedTransport, TTransportException)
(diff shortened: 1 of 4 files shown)


lekv:
@timarmstrong Do you have time to take a look at this one?

impala/_thrift_api.py
```diff
@@ -120,6 +123,43 @@ def get_socket(host, port, use_ssl, ca_cert):
...
+def get_http_transport(host, port, http_path, timeout=None, use_ssl=False,
+                       ca_cert=None, auth_mechanism='NOSASL', user=None,
+                       password=None):
+    # TODO: support timeout
+    if timeout is not None:
+        log.error('get_http_transport does not support a timeout')
+    if use_ssl:
+        url = 'https://%s:%s/%s' % (host, port, http_path)
+        log.debug('get_http_transport url=%s', url)
+        # TODO: Add server authentication with thrift 0.12
```
timarmstrong: Should probably raise an exception if ca_cert was not provided and we're not checking it

impala/hiveserver2.py
```diff
@@ -758,39 +759,47 @@ def threaded(func):
...
+            auth_mechanism=None, krb_host=None, use_http_transport=True,
+            http_path=''):
     log.debug('Connecting to HiveServer2 %s:%s with %s authentication '
               'mechanism', host, port, auth_mechanism)
-    sock = get_socket(host, port, use_ssl, ca_cert)
 
-    if krb_host:
-        kerberos_host = krb_host
+    if use_http_transport:
+        transport = get_http_transport(host, port, http_path=http_path,
```
timarmstrong: Do you need to pass in auth_mechanism?

timarmstrong: Validate that we don't have kerberos enabled for http?

impala/_thrift_api.py
lekv: Done, I added the check to the calling code because I didn't want to introduce dbapi errors to the thrift api wrapper.

impala/hiveserver2.py
lekv: Uh yes, fixed

lekv: Done

lekv:
Addressed the comments and tested the client using this change of Impala: https://gerrit.cloudera.org/#/c/14059/

I briefly looked into the LDAP test code but haven't had time to add a proper test.

impala/_thrift_api.py
```diff
@@ -120,6 +123,43 @@ def get_socket(host, port, use_ssl, ca_cert):
...
+        if user is None:
+            user = getpass.getuser()
+            log.debug('get_http_transport: user=%s', user)
+        if password is None:
+            if auth_mechanism == 'LDAP':
+                password = ''
+            else:
+                # PLAIN always requires a password for HS2.
+                password = 'password'
+            log.debug('get_http_transport: password=%s', password)
```
timarmstrong: It doesn't look like we use password, or am I missing something - it's all ldap_password below.

impala/hiveserver2.py
```diff
@@ -758,39 +759,55 @@ def threaded(func):
...
+            auth_mechanism=None, krb_host=None, use_http_transport=False,
+            http_path=''):
     log.debug('Connecting to HiveServer2 %s:%s with %s authentication '
               'mechanism', host, port, auth_mechanism)
-    sock = get_socket(host, port, use_ssl, ca_cert)
 
-    if krb_host:
-        kerberos_host = krb_host
+    if use_http_transport:
+        # TODO: Add server authentication with thrift 0.12
```
timarmstrong: Is there a thrift JIRA or something we can link to?

impala/_thrift_api.py
lekv: Fixed this one

impala/hiveserver2.py
lekv: Opened #362 and added it to the comment.

lekv:
Thanks for the comments. I added some LDAP tests to the Impala change that I used to test this: https://gerrit.cloudera.org/#/c/14059/

Once this is in, we will have to release a new version of Impyla, then we can bump the dependency in Impala and get that change in, too.

impala/hiveserver2.py
```diff
@@ -758,29 +759,46 @@ def threaded(func):
 
 def connect(host, port, timeout=None, use_ssl=False, ca_cert=None,
             user=None, password=None, kerberos_service_name='impala',
-            auth_mechanism=None, krb_host=None):
+            auth_mechanism=None, krb_host=None, use_http_transport=False,
```
dknupp: @lekv -- What is the difference between auth_mechanism=None and auth_mechanism=NOSASL? dbapi.py specifies these values:

`AUTH_MECHANISMS = ['NOSASL', 'PLAIN', 'GSSAPI', 'LDAP']`

NOSASL in the default in the connect() method from that file.

PR #363, opened 2019-09-03
fix ImportError  'THttpClient' in python3+

xyhXYH:
hi, My Python Version is  `python3.6`.

when `import hiveserver2`, then raise `ImportError: cannot import name 'THttpClient'`.

This is my solution, it worked.


xyhXYH:
I install impyla from source code.
```shell
git clone https://github.com/cloudera/impyla.git
cd impyla
python setup.py install
```

timarmstrong:
@xyhXYH Thank you for the fix!

PR #364, opened 2019-09-06
Set long_description_content_type to markdown in setup.py

dknupp:
PyPI now requires that the long description content type be specified.
(Default is reSturcturedText.)


dknupp:
@lekv @timarmstrong

PR #371, opened 2019-11-07
Update CHANGELOG for 0.16.1

dknupp:


dknupp:
@timarmstrong and @lekv

PR #373, opened 2019-11-11
add python 3.6 and change thrift_sasl to 0.4.0

Tagar:
add python 3.6 and change `thrift_sasl` to **0.4.0**

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ Required:
 
 Optional:
 
-* `thrift_sasl==0.2.1` for hive and/or Kerberos support:
+* `thrift_sasl==0.4.0` for hive and/or Kerberos support:
 
 * `pandas` for conversion to `DataFrame` objects; but see the [Ibis project][ibis] instead
 
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,7 @@ setup(
     extras_require={
         ":python_version>='3.0'": ["thriftpy2>=0.4.0,<0.5.0",
                                    ],
-        "kerberos": ["thrift_sasl==0.2.1",
+        "kerberos": ["thrift_sasl==0.4.0",
                     ],
     },
     keywords=('cloudera impala python hadoop sql hdfs mpp spark pydata '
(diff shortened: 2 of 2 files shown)


Tagar:
cc @dknupp and @timarmstrong 

please help to review this PR

Thank you!

setup.py
```diff
@@ -58,6 +58,7 @@ def readme():
         'Programming Language :: Python :: 3',
         'Programming Language :: Python :: 3.4',
```
Tagar: FYI - 

python 3.4 EOL'ed earlier this year - https://www.python.org/downloads/release/python-3410/ 

python 3.3 EOL'd in 2017 - https://www.python.org/downloads/release/python-337/

setup.py
```diff
@@ -43,9 +43,9 @@ def readme():
     package_data={'impala.thrift': ['*.thrift']},
     install_requires=['six', 'bitarray', 'thrift>=0.9.3'],
     extras_require={
-        ":python_version>='3.0'": ["thriftpy2>=0.4.0,<0.5.0",
+        ":python_version>='3.3'": ["thriftpy2>=0.4.0,<0.5.0",
```
Tagar: we could make this 3.4 to match below list of major Python 3 versions, or
add 3.3 below - please let me know what do you think.

thank you.

README.md
```diff
@@ -35,7 +35,7 @@ Required:
 
 Optional:
 
-* `thrift_sasl==0.2.1` for hive and/or Kerberos support:
+* `thrift_sasl==0.4.0` for hive and/or Kerberos support:
```
timarmstrong, line 38: Changed this to 0.4.2

setup.py
timarmstrong: I'm going to leave it as >= 3.0 just in case there are people using older releases of Python (sometimes we see weird things with extended support versions of RHEL, that kinda thing).

setup.py
timarmstrong: This makes sense, but we're pretty conservative with removing support for old Python versions because some of the linux vendors keep supporting them well past the upstream EOL.

timarmstrong:
I merged this with some other more recent changes - it's now in master - https://github.com/cloudera/impyla/commit/718106138bf28fd0d1dd6ad78e77776da6db3fb7

Issue #375, opened 2019-11-19
Remove usage of set literals


joemcdonnell:
Python 2.6 does not support set literals. Currently, interface.py has one location that uses a set literal:
https://github.com/cloudera/impyla/blob/86ca0d5507fcdaa0891b4d425ec81a22424332f5/impala/interface.py#L229
To maintain compatibility with Python 2.6, this can be converted to use a list:
`    if paramstyle in ('named', 'pyformat'):`
There may be other locations with set literals, so it would be useful to have some way to test syntax with Python 2.6.

timarmstrong:
I was able to run the tests in a centos 6 container with the fix.

`# On host:
git archive --prefix=impyla/ -o /tmp/impyla.tar FETCH_HEAD
docker run -v /tmp/impala.tar:/tmp/impala.tar -it centos:6  bash

# In container:
yum install -y epel-release
yum install -y python-pip pytest gcc python-devel
pip install six bitarray thrift unittest2

cd /tmp
tar xvf impyla.tar
cd impyla
export IMPYLA_TEST_HOST=<host ip>
py.test impala --verbose
`

I couldn't install sqlalchemy, so some tests failed, and other failed for other reasons, but looks the same otherwise as on an ubuntu 16. host.

`
[root@fa957af8e35f impala]# py.test impala --connect
================================================ test session starts =================================================
platform linux2 -- Python 2.6.6 -- pytest-2.4.2
collected 43 items / 1 errors 

impala/tests/test_data_types.py .
impala/tests/test_dbapi_compliance.py ...............F....F....F.........
impala/tests/test_hive.py .
impala/tests/test_query_parameters.py ......

======================================================= ERRORS =======================================================
__________________________________ ERROR collecting impala/tests/test_sqlalchemy.py __________________________________
impala/tests/test_sqlalchemy.py:17: in <module>
>   from sqlalchemy.engine import create_engine
E   ImportError: No module named sqlalchemy.engine
====================================================== FAILURES ======================================================
____________________________________________ ImpalaDBAPI20Test.test_close ____________________________________________

self = <impala.tests.test_dbapi_compliance.ImpalaDBAPI20Test testMethod=test_close>

    def test_close(self):
        con = self._connect()
        try:
            cur = con.cursor()
        finally:
            con.close()
    
        # cursor.execute should raise an Error if called after connection
        # closed
>       self.assertRaises(self.driver.Error,self.executeDDL1,cur)
[...]

PR #376, opened 2019-11-19
Remove usage of set literals #375

timarmstrong:
Testing:
Tested in a centos 6 VM.


timarmstrong:
I ran Impala tests using this version of the client on impala 2.6, and they all passed. So it appears to have fixed the issue.

PR #378, opened 2019-12-10
Make connecting over HTTP compatible with python 3.

dknupp:
- Under python 3, base64.encodestring() requires a byte sequence.

- For python 3, import THttpClient from thriftpy2.http

- Impyla uses thrift for python 2 and thriftpy2 for python 3.
  These libraries provide slightly different API's for opening the
  transport. For thrift, the method is called isOpen() regardless
  of the transport type. For thriftpy2, THttpClient retains isOpen(),
  but otherwise uses is_open().

Tested by connecting to unsecure, kerberized, and HTTP impalad's
using both python 2 and python 3.

### DIFF
--- a/impala/_thrift_api.py
+++ b/impala/_thrift_api.py
@@ -61,9 +61,9 @@ if six.PY2:
 
 
 if six.PY3:
-    # import thriftpy2 code
+    # When using python 3, import from thriftpy2 rather than thrift
     from thriftpy2 import load
-    from thrift.transport.THttpClient import THttpClient
+    from thriftpy2.http import THttpClient
     from thriftpy2.thrift import TClient, TApplicationException
     # TODO: reenable cython
     # from thriftpy2.protocol import TBinaryProtocol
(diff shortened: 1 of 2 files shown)


impala/_thrift_api.py
```diff
@@ -63,7 +63,7 @@
 if six.PY3:
     # import thriftpy2 code
     from thriftpy2 import load
```
timarmstrong: Should we always be using thriftpy2 instead of the thrift package in Python 3? Maybe leave a comment to that effect?

impala/hiveserver2.py
```diff
@@ -1034,9 +1034,19 @@ def _log_response(self, kind, response):
...
+    """
+    # python2 and thrift, or any THttpClient
+    if 'isOpen' in dir(transport):
+        transport_is_open = transport.isOpen()
+
+    # python3 and thriftpy2 (for TBufferedTransport only)
+    if 'is_open' in dir(transport):
+        transport_is_open = transport.is_open()
+
+    if not transport_is_open:
```
timarmstrong: What case is this needed for?

impala/_thrift_api.py
timarmstrong: Yep that would work

impala/hiveserver2.py
timarmstrong: Hm, ok, I guess we can leave it alone.

timarmstrong:
OK to merge once that comment is added.

PR #388, opened 2020-03-13
[Fix #351 Improves the example of a cursor iteration to execute the `print` fun…

### LINKED ISSUE #351: Update README to import process
I'm new to this package and was attempting some of the examples in the README, and am finding myself very confused. Not sure if it's the code or the explanation.

Lines 107 - 111 have the following;
```python
cursor.execute('SELECT * FROM mytable LIMIT 100')
for row in cursor:
    process(row)
```
But the `process` function is not imported anywhere. 

Is this an actual function? - If so where is it from / what does it do? 

Or, is this psuedo for 'do some process'? - If so it would be good to explain this in the wording around the code chunk just for those new to the package who spend a while looking for the `process` function: "The `Cursor` object also exposes the iterator interface, which is buffered
(controlled by `cursor.arraysize`), so any 'process' type function can be used on the values:"

Apologies if this is really simplistic and obvious to you - but I got confused while looking at it, so thought it would be good to let you know!

Thanks

rafaelreuber:
Improves the example of a cursor iteration to execute the `print` function instead `process`. 
This could be a bit confusing for inexperienced users. [Fix #351

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ The `Cursor` object also exposes the iterator interface, which is buffered
 ```python
 cursor.execute('SELECT * FROM mytable LIMIT 100')
 for row in cursor:
-    process(row)
+    print(row)
 ```
 
 Furthermore the `Cursor` object returns you information about the columns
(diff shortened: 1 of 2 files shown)


timarmstrong:
Thanks for your contribution!

PR #390, opened 2020-03-18
Add the util function as_dict 

rafaelreuber:
Add the util function as_dict. This is very useful to create APIs.

### DIFF
--- a/impala/util.py
+++ b/impala/util.py
@@ -64,6 +64,18 @@ def as_pandas(cursor, coerce_float=False):
                                   coerce_float=coerce_float)
 
 
+def as_dict(cursor):
+    """
+    :param cursor: `HiveServer2Cursor`
+        The cursor object that has a result set waiting to be fetched.
+    :return: list of dict
+    """
+    columns = [col[0].lower() for col in cursor.description]
+    data = [
+        dict(zip(columns, row)) for row in cursor.fetchall()]
+    return data
+
+
 def _random_id(prefix='', length=8):
     return prefix + ''.join(random.sample(string.ascii_uppercase, length))
 
--- a/README.md
+++ b/README.md
@@ -127,7 +127,8 @@ with open(targetfile, 'w', newline='') as outcsv:
         writer.writerow(row)
 ```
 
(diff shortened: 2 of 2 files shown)


impala/util.py
```diff
@@ -64,6 +64,18 @@ def as_pandas(cursor, coerce_float=False):
                                   coerce_float=coerce_float)
 
 
+def as_dict(cursor):
```
timarmstrong: I think as_dicts() would be a clearer name (it sounds like it returns a single dictionary).

timarmstrong: Can you add a test for this method. E.g. see impala/tests/test_hive_dict_cursor.py

rafaelreuber: You are right! Another option is `dictfetchall`. This name refers to the fetchall operation that occurs on the function.

rafaelreuber:
Hi timarmstrong,

What you think about both `as_dicts` and `as_pandas` as methods from HiveServer2Cursor. 

```python
class HiveServer2Cursor(Cursor):

    def as_dicts(self):
        pass
    def as_pandas(self):
        pass
```

I think this provides a concise interface.

```python
with conn.cursor() as cursor:
    cursor.execute('SELECT * FROM mytable LIMIT 100')
    data = cursor.as_dict()

with conn.cursor() as cursor:
    cursor.execute('SELECT * FROM mytable LIMIT 100')
    df = cursor.as_pandas()	
```

And of course the class HiveServer2DictCursor could be removed.

timarmstrong:
I think those are both useful methods. My ask is to not remove any existing functionality that people may depend on, but adding new convenience methods seems like a good idea.

dknupp:
@rafaelreuber @timarmstrong -- Does adding an `as_pandas()` method imply that pandas is being added as a library dependency?

timarmstrong:
We already have an as_pandas method, but it only imports pandas inside the method. So I guess it's only a dependency if you want to use that method.

impala/util.py
timarmstrong: Either is fine by me. It looks like the dictfetchall is used by django so there's some precedent.

Issue #396, opened 2020-04-30
Impyla doesn't work against Hive client protocol version >=10


attilajeges:
Impyla still uses Hive client version V6 and doesn't work correctly against servers supporting Hive client protocol version >= V10.

The issue happens when async compilation is enabled. Impyla correctly waits for the query to get into a fetchable state. After that, it checks TOperationHandle.hasResultSet before trying to fetch the results.
The problem is that TOperationHandle.hasResultSet is not set correctly any longer. The flag was moved to TGetOperationStatusResp in later Hive client versions.

attilajeges:
The fix has been merged in

PR #397, opened 2020-05-01
Fix interoperability with Hive client version >= V10 (#396)

attilajeges:
Impyla uses Hive client version V6 and doesn't work correctly
against Hive client protocol version >= V10.

The issue happens when async compilation is enabled:
1. Impyla waits for the query to get into a fetchable state.
2. Impyla checks TOperationHandle.hasResultSet before trying to
fetch the results.

The problem is that TOperationHandle.hasResultSet is not set
correctly any longer. The flag was moved to
TGetOperationStatusResp in later Hive client versions.

This fix makes sure that TGetOperationStatusResp.hasResultSet is
used instead of TOperationHandle.hasResultSet if present.

Testing:
- Manually tested against a Hive client version V10.

### DIFF
--- a/impala/_thrift_gen/TCLIService/ttypes.py
+++ b/impala/_thrift_gen/TCLIService/ttypes.py
@@ -5626,6 +5626,7 @@ class TGetOperationStatusResp(object):
    - sqlState
    - errorCode
    - errorMessage
+   - hasResultSet
   """
 
   thrift_spec = (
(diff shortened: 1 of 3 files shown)


dknupp:
Thanks for the quick turnaround on this issue @attilajeges -- this definitely goes beyond my understanding of thrift and hive, so I'm glad you took it on. I guess my first naive question is this. You're making changes to files here that are thrift generated (like ttypes.py). When we officially update to a later version of thrift, will these changes need to be applied again? Or will they already be included?

timarmstrong:
This LGTM as a targeted fix.

attilajeges:
> Thanks for the quick turnaround on this issue @attilajeges -- this definitely goes beyond my understanding of thrift and hive, so I'm glad you took it on. I guess my first naive question is this. You're making changes to files here that are thrift generated (like ttypes.py). When we officially update to a later version of thrift, will these changes need to be applied again? Or will they already be included?

They will be included.

Issue #380, opened 2019-12-20
Report http error codes correctly when using hs2-http


bartash:
In Impala shell 'IMPALA-9240 Impala shell using hs2-http reports all http error codes as EOFError' deals with the case where an http call fails when sending a thrift request to Impala. I think the same problem exists in Impyla. Please backport the fix for IMPALA-9240 to Impyla

PR #400, opened 2020-05-15
Add HTTP code handling to THttpClient (#380)

bartash:
Before this change Impyla is not checking HTTP return codes when
using the hs2-http protocol. Impyla is sending a request message
(e.g. send_CloseOperation) but the HTTP call to send this message may
fail. This will result in a failure when reading the reply (e.g. in
recv_CloseOperation) as there is no reply data to read. This will
typically result in an 'EOFError'.

Replace use of HTTPClient with ImpalaHttpClient. This code is copied
from Impala Shell. Make some small changes to enable this code to run on
Python3 as well as Python2. Http errors result in the throwing of a new
exception 'HttpError' (a subclass of RpcError) so that the http code can
be accessed directly.

This change does not contain any attempt to recover from an HTTP
failures but it does allow the failure to be detected. In future it may
be possible to retry after certain HTTP errors.

Testing:
- Add a new test that tries to connect to an HTTP
  server that always returns a 503 error. Check that an HttpError
  exception is thrown that contains the Http error code and message.

### DIFF
--- a/impala/_thrift_api.py
+++ b/impala/_thrift_api.py
@@ -23,10 +23,17 @@ from __future__ import absolute_import
 import base64
 import getpass
 import os
+from io import BytesIO
+
+from six.moves import urllib
+from six.moves import http_client
+import warnings
+
 import six
 import ssl
 import sys
 
+from impala.error import HttpError
 from impala.util import get_logger_and_init_null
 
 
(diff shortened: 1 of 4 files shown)


impala/_thrift_api.py
```diff
@@ -103,6 +108,184 @@
     ThriftClient = TClient
 
 
+class ImpalaHttpClient(TTransportBase):
+  """Http implementation of TTransport base."""
```
timarmstrong, line 112: Maybe leave a comment with a pointer to indicate it's a copy of the code from impala shell?

bartash: Thanks,I did this but I added a small change

PR #415, opened 2020-09-03
Implement GSSAPI authentication over http transport.

attilajeges:
The implementation adds a new dependency on 'kerberos' python module,
which is a pip-installed module distributed under Apache License Version 2.

When using impyla with GSSAPI over http, it is assumed that the client has a
preexisting kinit-cached Kerberos ticket that impyla can pass to the server
automatically without the user needing to reenter their password.

Examples:

1. Hive
conn = connect('hiveserver2.example.com', 10001,
  auth_mechanism="GSSAPI",
  kerberos_service_name='hive',
  use_http_transport=True,
  http_path='cliservice',
  auth_cookie_name='hive.server2.auth')
cursor = conn.cursor()
cursor.execute("show databases")
res = cursor.fetchall()

2. Impala
conn = connect('impala-coordinator.exmple.com', 28000,
  auth_mechanism="GSSAPI",
  kerberos_service_name='impala',
  use_http_transport=True,
  http_path='cliservice',
  auth_cookie_name='impala.auth')
cursor = conn.cursor()
cursor.execute("show databases")
res = cursor.fetchall()

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ Optional:
 * `thrift_sasl>=0.2.1` for hive and/or Kerberos support. This also requires a SASL
    library to be installed on your system - see [System SASL](#system-sasl)
 
-* `kerberos` for Kerberos over HTTP support.
+* `kerberos>=1.3.0` for Kerberos over HTTP support.
 
 * `pandas` for conversion to `DataFrame` objects; but see the [Ibis project][ibis] instead
 
--- a/README.md
+++ b/README.md
@@ -40,6 +40,8 @@ Optional:
 * `thrift_sasl>=0.2.1` for hive and/or Kerberos support. This also requires a SASL
    library to be installed on your system - see [System SASL](#system-sasl)
 
+* `kerberos` for Kerberos over HTTP support.
+
 * `pandas` for conversion to `DataFrame` objects; but see the [Ibis project][ibis] instead
 
 * `sqlalchemy` for the SQLAlchemy engine
--- a/impala/_thrift_api.py
+++ b/impala/_thrift_api.py
@@ -272,6 +272,9 @@ class ImpalaHttpClient(TTransportBase):
       self.__auth_cookie = None
     return self.__auth_cookie
 
+  def isAuthCookieSet(self):
+    return self.__auth_cookie is not None
+
   def deleteAuthCookie(self):
     self.__auth_cookie = None
     self.__auth_cookie_expiry = None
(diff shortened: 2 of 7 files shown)


impala/_thrift_api.py
```diff
@@ -120,17 +122,19 @@ class ImpalaHttpClient(TTransportBase):
...
     ImpalaHttpClient(host, port, path) - deprecated
     ImpalaHttpClient(uri, [port=<n>, path=<s>, cafile=<filename>, cert_file=<filename>,
-        key_file=<filename>, ssl_context=<context>])
+        key_file=<filename>, ssl_context=<context>, auth_cookie_name=<cookiename>])
 
     Only the second supports https.  To properly authenticate against the server,
     provide the client's identity by specifying cert_file and key_file.  To properly
     authenticate the server, specify either cafile or ssl_context with a CA defined.
     NOTE: if both cafile and ssl_context are defined, ssl_context will override cafile.
+    auth_cookie_name is used to specify the name of the cookie used for cookie-based
```
timarmstrong, line 136: It seems inconvenient to require specifying this manually. Can we default it to 'impala.auth' at least? Or detect it automatically in setAuthCookie. E.g. if __auth_cookie_name is None, then check to see if "impala.auth' or 'hive.server2.auth' is present in the response and then use that?

README.md
```diff
@@ -40,6 +40,8 @@ Optional:
 * `thrift_sasl>=0.2.1` for hive and/or Kerberos support. This also requires a SASL
    library to be installed on your system - see [System SASL](#system-sasl)
 
+* `kerberos` for Kerberos over HTTP support.
```
timarmstrong, line 43: What version of the kerberos module does this require? OK if it's just the most recent release but we should document.

What Python versions did you test with? Would be good to test with python 2.7 and a recent Python 3 version at least.

timarmstrong (changes requested):

README.md
myloginid: We should at minimum test with with CDSW/CML Python 2 and 3 versions that are listed below. This is to ensure that this feature works seemlessly in CDSW/CML
https://docs.cloudera.com/documentation/data-science-workbench/latest/topics/cdsw_release_notes.html#concept_mxd_jmg_1kb__section_e14_hcr_jkb

impala/_thrift_api.py
attilajeges: 'auth_cookie_name' was renamed to 'auth_cookie_names' and now it takes a list of cookie names to look for.  The default value (specified in connect() API call) is ['impala.auth', 'hive.server2.auth'] so impyla will detect these by default.

README.md
attilajeges: I've tested the patch with python 2.6.9, 2.7.17 and 3.6.9.

I've only tested it with the most recent version of the kerberos module, which is 1.3.0, so I've added 1.3.0 to the README file and to setup.py.

timarmstrong:
Thanks for the contribution @attilajeges

PR #419, opened 2020-09-30
Fix #418, no 'message' attribute so just cast to str to extract message

### LINKED ISSUE #418: If sasl is missing, impyla becomes incompatible with python3
This line is python2-only as `message` is not an attribute anymore:
https://github.com/cloudera/impyla/blob/master/impala/sasl_compat.py#L10

timkpaine:


timarmstrong (approved):

timarmstrong:
Thanks for the contribution!

PR #424, opened 2020-12-22
Implement simple retry which throws the underlying HttpError if retrying fails

bartash:
Copy design from IMPALA-9466 and add client retry for hs2-http protocol.
Added retries for idempotent rpcs: OpenSession, GetResultSetMetadata,
CancelOperation, GetOperationStatus, GetRuntimeProfile, GetExecSummary,
and GetLog. Not done were PingImpalaHS2Service (not available) and
CloseImpalaOperation (as we cannot distinguish the non-dml statements
which are idempotent) Ported test_hs2_fault_injection.py from Impala
shell.

Add http_headers to HttpError and sleep if the 'Retry-After' header
contains an integer value.

Allow override of the http_port by setting IMPYLA_TEST_HTTP_PORT

All tests ran clean with Python2 and Python3.

### DIFF
--- a/impala/hiveserver2.py
+++ b/impala/hiveserver2.py
@@ -14,25 +14,21 @@
 
 from __future__ import absolute_import
 
-import six
-import time
 import getpass
-import datetime
-import socket
-import operator
 import re
-import sys
-from six.moves import range
-from bitarray import bitarray
+import socket
 
-from impala.compat import Decimal
-from impala.util import get_logger_and_init_null
-from impala.interface import Connection, Cursor, _bind_parameters
-from impala.error import (NotSupportedError, OperationalError,
-                          ProgrammingError, HiveServer2Error)
+import datetime
+import operator
+import six
+import sys
+import time
+from bitarray import bitarray
+from six.moves import range
+
 from impala._thrift_api import (
-    get_socket, get_http_transport, get_transport, ImpalaHttpClient,
-    TTransportException, TBinaryProtocol, TOpenSessionReq, TFetchResultsReq,
+    get_socket, get_http_transport, get_transport, TTransportException, TBinaryProtocol, TOpenSessionReq,
+    TFetchResultsReq,
     TCloseSessionReq, TExecuteStatementReq, TGetInfoReq, TGetInfoType, TTypeId,
     TFetchOrientation, TGetResultSetMetadataReq, TStatusCode, TGetColumnsReq,
     TGetSchemasReq, TGetTablesReq, TGetFunctionsReq, TGetOperationStatusReq,
@@ -40,7 +36,11 @@ from impala._thrift_api import (
     TProtocolVersion, TGetRuntimeProfileReq, TRuntimeProfileFormat,
     TGetExecSummaryReq, ImpalaHiveServer2Service, TExecStats, ThriftClient,
     TApplicationException)
-
+from impala.compat import Decimal
+from impala.error import (NotSupportedError, OperationalError,
+                          ProgrammingError, HiveServer2Error, HttpError)
+from impala.interface import Connection, Cursor, _bind_parameters
+from impala.util import get_logger_and_init_null
 
 log = get_logger_and_init_null(__name__)
 
@@ -402,7 +402,7 @@ class HiveServer2Cursor(Cursor):
         loop_start = time.time()
         while True:
             req = TGetOperationStatusReq(operationHandle=self._last_operation.handle)
-            resp = self._last_operation._rpc('GetOperationStatus', req)
+            resp = self._last_operation._rpc('GetOperationStatus', req, True)
             self._last_operation.update_has_result_set(resp)
             operation_state = TOperationState._VALUES_TO_NAMES[resp.operationState]
 
@@ -796,7 +796,7 @@ def threaded(func):
 def connect(host, port, timeout=None, use_ssl=False, ca_cert=None,
             user=None, password=None, kerberos_service_name='impala',
             auth_mechanism=None, krb_host=None, use_http_transport=False,
-            http_path='', auth_cookie_names=None):
+            http_path='', auth_cookie_names=None, retries=3):
     log.debug('Connecting to HiveServer2 %s:%s with %s authentication '
               'mechanism', host, port, auth_mechanism)
 
@@ -847,7 +847,7 @@ def connect(host, port, timeout=None, use_ssl=False, ca_cert=None,
     log.debug('transport=%s protocol=%s service=%s', transport, protocol,
               service)
 
-    return HS2Service(service)
+    return HS2Service(service, retries=retries)
 
 
 def _is_columnar_protocol(hs2_protocol_version):
@@ -1003,18 +1003,19 @@ class ThriftRPC(object):
         self.client = client
         self.retries = retries
 
-    def _rpc(self, func_name, request):
+    def _rpc(self, func_name, request, retry_on_http_error=False):
         self._log_request(func_name, request)
-        response = self._execute(func_name, request)
+        response = self._execute(func_name, request, retry_on_http_error)
         self._log_response(func_name, response)
         err_if_rpc_not_ok(response)
         return response
 
-    def _execute(self, func_name, request):
+    def _execute(self, func_name, request, retry_on_http_error=False):
         # pylint: disable=protected-access
         # get the thrift transport
         transport = self.client._iprot.trans
         tries_left = self.retries
+        last_http_exception = None
         while tries_left > 0:
             try:
                 log.debug('Attempting to open transport (tries_left=%s)',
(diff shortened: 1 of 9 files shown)


impala/hiveserver2.py
```diff
@@ -1026,20 +1027,47 @@ def _execute(self, func_name, request):
...
                               tries_left)
+                last_http_exception = None
+            except HttpError as h:
+                if not retry_on_http_error:
+                    log.debug('Caught HttpError %s %s in %s which is not retryable',
+                              h, str(h.body or ''), func_name)
+                    raise
+                last_http_exception = h
+                if tries_left > 0:
+                    retry_secs = None
```
abhir26, line 1042: Should we add a default sleep interval such as 1 sec and use exponential back off and progressively increase the wait interval for every retry attempt? If a failure occurred then retrying instantly would most probably result in failure again.
For the case where the http header has the retry after header we could follow the sleep duration from the header.

impala/hiveserver2.py
```diff
@@ -1112,37 +1143,43 @@ def __init__(self, service, handle, config, hs2_protocol_version,
 
     def close(self):
         req = TCloseSessionReq(sessionHandle=self.handle)
-        self._rpc('CloseSession', req)
+        # CloseSession rpcs don't need retries since we catch all exceptions and close
```
abhir26, line 1146: The close function doesn't seem to handle exceptions. Are we catching the exceptions and closing the transport in a top level caller somewhere?

impala/hiveserver2.py
```diff
@@ -1112,37 +1143,43 @@ def __init__(self, service, handle, config, hs2_protocol_version,
...
+        # Do not try to retry http requests.
+        # Read queries should be idempotent but most dml queries are not. Also retrying
+        # query execution from client could be expensive and so likely makes sense to do
+        # it if server is also aware of the retries.
+        return self._operation('ExecuteStatement', req, False)
 
     def get_databases(self, schema='.*'):
         req = TGetSchemasReq(sessionHandle=self.handle, schemaName=schema)
-        return self._operation('GetSchemas', req)
+        return self._operation('GetSchemas', req, False)
```
abhir26, line 1163: Typically all the read operations such as GetSchemas, GetTables, GetColumns, GetFunctions should be idempotent, as long as they return all the results in a single call - it's either all or nothing and so retrying these functions should be okay.

But, I am okay if you prefer to limit the scope of retries to the more common functions at this time.

impala/hiveserver2.py
```diff
@@ -1112,37 +1143,43 @@ def __init__(self, service, handle, config, hs2_protocol_version,
...
         req = TExecuteStatementReq(sessionHandle=self.handle,
                                    statement=statement,
                                    confOverlay=configuration,
                                    runAsync=run_async)
-        return self._operation('ExecuteStatement', req)
+        # Do not try to retry http requests.
+        # Read queries should be idempotent but most dml queries are not. Also retrying
+        # query execution from client could be expensive and so likely makes sense to do
+        # it if server is also aware of the retries.
+        return self._operation('ExecuteStatement', req, False)
```
abhir26, line 1159: Thanks for explicitly passing a **False** param even though the default value is False for retries. Improves the readability.

impala/tests/test_hs2_fault_injection.py
```diff
@@ -0,0 +1,357 @@
...
+        except HttpError as e:
+            assert str(e) == 'HTTP code 502: Injected Fault'
+        assert query_handle is None
+        cur.close()
+        con.close()
+        assert self.__expect_msg_no_retry("ExecuteStatement") in caplog.text
+
+    def test_get_query_state(self, caplog):
+        """Tests fault injection in fetchall().
+        GetOperationStatus rpc fails and results in error since retries are not supported."""
```
abhir26, line 276: GetOperationStatus is idempotent. fetchall() executes 'GetOperationStatus' and 'Fetch' in a loop untill all the data is fetched. This test should be successful with retry.

impala/tests/test_hs2_fault_injection.py
```diff
@@ -0,0 +1,357 @@
...
+        cur.execute('select 1', {})
+        self.transport.enable_fault(502, "Injected Fault", 0.1)
+        cur.fetchcbatch()
+        cur.close()
+        con.close()
+        assert self.__expect_msg_retry("GetResultSetMetadata") in caplog.text
+
+    def test_fetch_results(self, caplog):
+        """Tests fault injection in fetchcbatch().
+        GetResultSetMetadata rpc fails and is retried succesfully."""
```
abhir26, line 302: The comment probably needs to be updated.

impala/tests/test_hs2_fault_injection.py
```diff
@@ -0,0 +1,357 @@
...
+        except HttpError as e:
+            assert str(e) == 'HTTP code 502: Injected Fault'
+        self.transport.disable_fault()
+        cur.close()
+        con.close()
+        assert self.__expect_msg_no_retry("FetchResults") in caplog.text
+
+    def test_close_operation(self, caplog):
+        """Tests fault injection in fetchcbatch().
+        GetResultSetMetadata rpc fails and is retried successfully."""
```
abhir26, line 320: Comment needs to be updated.

abhir26 (approved):

impala/hiveserver2.py
bartash: I'll change the code for these 4 functions

impala/hiveserver2.py
bartash: I'm adding a 1 second sleep

impala/tests/test_hs2_fault_injection.py
bartash: Thanks - updating docstring comment

impala/tests/test_hs2_fault_injection.py
bartash: Thanks

impala/tests/test_hs2_fault_injection.py
bartash: Thanks

impala/hiveserver2.py
bartash: There is no top level handler so the comment is a lie (copied from shell).
I think CloseSession would be non-idempotent as you cannot close a session twice
https://github.com/apache/impala/blob/4099a606892c377b9e8c9c6df2a45a7d42afcaea/tests/hs2/test_hs2.py#L203
I will fix the comment but I think closing the transport is a non-trivial change.

impala/hiveserver2.py
```diff
@@ -1003,18 +1003,19 @@ def __init__(self, client, retries=3):
...
         return response
 
-    def _execute(self, func_name, request):
+    def _execute(self, func_name, request, retry_on_http_error=False):
         # pylint: disable=protected-access
         # get the thrift transport
         transport = self.client._iprot.trans
         tries_left = self.retries
+        last_http_exception = None
         while tries_left > 0:
```
attilajeges, line 1019: Not this change, but I think this condition should be "tries_left >= 0" instead.
(I assume self.retries == 0 means the call should be attempted once but not retried after a failure).

bartash: I think self.retries == 1 means that the call should be attempted once without retries
I agree that you could argue that this could be self.tries as when it is 1 there is no actual 'retry'.
I am going to leave the naming like this for now.
But I am making the change you suggest on L1041, thank you.

bartash:
Can you take a quick look @attilajeges  before I push as abhir26 is away from computers right now?

impala/hiveserver2.py
```diff
@@ -1214,54 +1253,62 @@ def update_has_result_set(self, state):
...
     def cancel(self):
         req = TCancelOperationReq(operationHandle=self.handle)
-        return self._rpc('CancelOperation', req)
+        # CancelOperation rpc is idempotent and so safe to retry.
+        return self._rpc('CancelOperation', req, True)
 
     def close(self):
         req = TCloseOperationReq(operationHandle=self.handle)
-        return self._rpc('CloseOperation', req)
+        # CloseImpalaOperation rpc is not idempotent for dml and we're not sure
```
attilajeges, line 1293: nit: The comment should be about "CloseOperation".

impala/tests/test_hs2_fault_injection.py
```diff
@@ -0,0 +1,364 @@
...
+        return ("Caught HttpError HTTP code 503: Injected Fault  in {0} (tries_left=3), retry after 1 secs".
+                format(impala_rpc_name))
+
+    def __expect_msg_no_retry(self, impala_rpc_name):
+        """Returns expected log message for rpcs which can not be retried"""
+        return ("Caught HttpError HTTP code 502: Injected Fault  in {0} which is not retryable".
+                format(impala_rpc_name))
+
+    def test_connect(self, caplog):
+        """Tests fault injection in connect().
```
attilajeges, line 177: I could be wrong here, but I think the fault is injected in cursor() call. OpenSession rpc is called from cursor().

impala/tests/test_hs2_fault_injection.py
```diff
@@ -0,0 +1,364 @@
...
+        caplog.set_level(logging.DEBUG)
+        self.transport.enable_fault(502, "Injected Fault", 0.2)
+        con = self.connect()
+        cur = con.cursor(configuration=self.configuration)
+        cur.close()
+        con.close()
+        assert self.__expect_msg_retry("OpenSession") in caplog.text
+
+    def test_connect_proxy(self, caplog):
+        """Tests fault injection in connect().
```
attilajeges, line 189: Same here and below.

impala/tests/util.py
```diff
@@ -42,6 +43,11 @@ def __init__(self, host=None, port=None, hive_port=None, auth_mech=None):
         else:
             self.port = get_env_var('IMPYLA_TEST_PORT', int, 21050)
 
+        if http_port is not None:
+            self.http_port = port
```
attilajeges, line 47: I think, this should be: 
self.http_port = http_port

impala/hiveserver2.py
bartash: done

impala/tests/test_hs2_fault_injection.py
bartash: You're right. Done.

impala/tests/test_hs2_fault_injection.py
bartash: done

impala/tests/util.py
bartash: yes good catch

PR #433, opened 2021-03-11
Server certs should not be verified if SSL is enabled and ca_cert is not specified

attilajeges:
The docstring for connect() method in impala.dbapi module states:

ca_cert : str, optional
  Local path to the the third-party CA certificate. If SSL is enabled
  but the certificate is not specified, the server certificate will
  not be validated.

However, the current behavior is different: if SSL is enabled,
'ca_cert' is set to None and HTTP transport is used, impyla will still
check the server certificate and reject it if it is invalid (or
self-signed).

This behavior is incorrect and not in line with impala-shell.

Testing:
- All tests ran clean with Python2 and Python3.
- Manually tested against impala coordinator with self-signed server
  certificate.


bartash (approved):
LGTM

PR #437, opened 2021-04-08
Added "fetchType" to TFetchResultsReq.

scarlin-cloudera:
The "fetchType" attribute was removed from an earlier commit. There
is code that is still reference the fetchType in hiveserver2.py. The
code was in an exception so it was most likely never hit. However, it
is needed for any HS2 api that has not implemented the deprecated
GetLog API call.


attilajeges (approved):

bartash (approved):
LGTM

PR #439, opened 2021-04-14
Unify Python 2 and 3 thrift handling and remove thriftpy2

csringhofer:
Apache Thrift supports Python 3 since 0.10.0 so switching to 0.11.0
allowed to use the same implementation for both Python versions.

Before this commit thriftpy2 was used with Python 3. Now thriftpy2
and the hacks to hide the differences between the two Thrift libraries
are removed.

### DIFF
--- a/impala/_thrift_api.py
+++ b/impala/_thrift_api.py
@@ -48,9 +48,6 @@ log = get_logger_and_init_null(__name__)
 from thrift.transport.TSocket import TSocket
 from thrift.transport.TTransport import (
     TBufferedTransport, TTransportException, TTransportBase)
-from thrift.Thrift import TApplicationException
-from thrift.protocol.TBinaryProtocol import (
-    TBinaryProtocolAccelerated as TBinaryProtocol)
 
 # import HS2 codegen objects
 from impala._thrift_gen.ImpalaService import ImpalaHiveServer2Service
--- a/impala/_thrift_api.py
+++ b/impala/_thrift_api.py
@@ -12,9 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# This package is here to clean up references to thrift, because we're using
-# thriftpy2 for Py3 at the moment.  This should all be temporary, as Apache
-# Thrift gains Py3 compatibility.
+# This package's main goal in the past was to clean up references to thrift, because
+# we were using thriftpy2 for Py3. This is no longer necessary since upgrading to
+# Thrift 0.11.0, as Thrift supports Python 3 since 0.10.0. Now there are only some
+# leftover utility classes and functions.
 
 # pylint: disable=wrong-import-position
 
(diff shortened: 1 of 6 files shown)


impala/_thrift_api.py
```diff
@@ -42,74 +43,19 @@
...
-    from impala.RuntimeProfile_thrift import TRuntimeProfileFormat
-    ThriftClient = TClient
+# pylint: disable=import-error,unused-import
+# import Apache Thrift code
+from thrift.transport.TSocket import TSocket
+from thrift.transport.TTransport import (
+    TBufferedTransport, TTransportException, TTransportBase)
+from thrift.Thrift import TApplicationException
+from thrift.protocol.TBinaryProtocol import (
+    TBinaryProtocolAccelerated as TBinaryProtocol)
```
attilajeges, line 53: Minor: I think 'TBinaryProtocolAccelerated' was renamed to 'TBinaryProtocol' because in 'thriftpy2.protocol.binary' 'TBinaryProtocol' was the corresponding type name. Probably it would be more straightforward to use the original name 'TBinaryProtocolAccelerated' now on, since thriftpy2 is not used anymore.

impala/_thrift_api.py
```diff
@@ -42,74 +43,19 @@
...
+
+# import HS2 codegen objects
+from impala._thrift_gen.ImpalaService import ImpalaHiveServer2Service
+ThriftClient = ImpalaHiveServer2Service.Client
```
attilajeges, line 57: Minor: Again, introducing a new name for 'ImpalaHiveServer2Service.Client' is probably not necessary anymore. I'd just use 'ImpalaHiveServer2Service.Client' if it is not too much trouble.

impala/tests/test_hs2_fault_injection.py
```diff
@@ -15,23 +15,15 @@
 
 import six
 
-if six.PY2:
-    from thrift.protocol.TBinaryProtocol import (
-        TBinaryProtocolAccelerated as TBinaryProtocol)
-    # noinspection PyProtectedMember
-    from impala._thrift_gen.ImpalaService import ImpalaHiveServer2Service
+from thrift.protocol.TBinaryProtocol import (
+    TBinaryProtocolAccelerated as TBinaryProtocol)
```
attilajeges, line 19: Again, no need to rename 'TBinaryProtocolAccelerated' to 'TBinaryProtocol'

impala/tests/test_hs2_fault_injection.py
```diff
@@ -15,23 +15,15 @@
...
 
-    ThriftClient = ImpalaHiveServer2Service.Client
-
-if six.PY3:
-    # When using python 3, import from thriftpy2 rather than thrift
-    # TODO: reenable cython
-    from thriftpy2.protocol.binary import TBinaryProtocol  # noqa
-    from thriftpy2.transport import (TSocket, TTransportException, TTransportBase)  # noqa
-    from thriftpy2.transport.buffered import TBufferedTransport  # noqa
+ThriftClient = ImpalaHiveServer2Service.Client
```
attilajeges, line 23: Same here

impala/_thrift_api.py
csringhofer: Actually it was not used in this file at all.

PR #441, opened 2021-04-22
Extensions to SQLAlchemy ImpalaDLLCompiler to support Alembic schema migrations

mrdavidlaing:
Specifically, 

1. Extend ImpalaDLLCompiler to prevent CONSTRAINT and NOT NULL clauses being added to CreateTable DDL and 
2. Translate all "string" types to STRING and "datetime" types to TIMESTAMP.

This prevents the `alembic_version` table DDL from containing a primary key constraint and a NOT NULL VARCHAR column that aren't supported in an Impala TEXT table.

### DIFF
--- a/impala/sqlalchemy.py
+++ b/impala/sqlalchemy.py
@@ -48,6 +48,21 @@ class STRING(String):
 
 
 class ImpalaDDLCompiler(DDLCompiler):
+    # Impala has no support for foreign keys.
+    def visit_foreign_key_constraint(self, constraint):
+        return None
+
+    # Impala has no support for primary keys.
+    def visit_primary_key_constraint(self, constraint):
+        return None
+
+    def get_column_specification(self, column, **kwargs):
+        column.nullable = True  # Prevent adding NOT NULL constraints; since all columns in Impala are nullable
+        colspec = super(ImpalaDDLCompiler, self).get_column_specification(
+            column, **kwargs
+        )
+        return colspec
+
     def post_create_table(self, table):
         """Build table-level CREATE options."""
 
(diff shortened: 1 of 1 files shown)


impala/sqlalchemy.py
```diff
@@ -70,6 +88,23 @@ def post_create_table(self, table):
 class ImpalaTypeCompiler(GenericTypeCompiler):
     # pylint: disable=unused-argument
 
+    # https://docs.cloudera.com/documentation/enterprise/6/6.3/topics/impala_porting.html
+    # Impala only supports the STRING type
```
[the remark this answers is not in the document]
mrdavidlaing: Thanks for pointing this out.

I'm unsure if you're suggesting a change here.  I'm happy to make one - just unsure what that change might be.

impala/sqlalchemy.py
```diff
@@ -70,6 +88,26 @@ def post_create_table(self, table):
...
+
+    # Impala only supports the TIMESTAMP type
+    def visit_DATETIME(self, type_):
+        return 'TIMESTAMP'
+
+    visit_TIME = visit_DATETIME
+
+    # Impala > 3.4.0 support the DATE type - https://issues.apache.org/jira/browse/IMPALA-6169
+    def visit_DATE(self, type_):
+        return 'DATE'
```
mrdavidlaing, line 109: @csringhofer Is there a way to "detect" the version of Impala being used within the context of this method?

If so; I'm thinking we could do something like:

```
if _impala_version > 3.4.0:
  return 'DATE'
else:
  return 'TIMESTAMP'
```

impala/sqlalchemy.py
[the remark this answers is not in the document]
csringhofer: I am unsure about what is the best to do here - the safest seems to be leaving CHAR and VARCHAR unchanged.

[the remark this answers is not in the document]
mrdavidlaing: This should be resolved by 1179f1e

impala/sqlalchemy.py
csringhofer: We create the ImpalaTypeCompiler here: https://github.com/cloudera/impyla/blob/master/impala/sqlalchemy.py#L159
At that point we haven't connected to Impala yet, so we cannot know the exact version. I am also concerned making such behavior version dependent, because it makes it possible for an Impala version upgrade to brake an existing workflow.

A possible solution is to register two dialects in the future, e.g. impala and impala4, and impala4 could contain logic that only works on newer Impala versions.

mrdavidlaing: > A possible solution is to register two dialects in the future, e.g. impala and impala4, and impala4 could contain logic that only works on newer Impala versions.

I like this idea.  I'll modify the PR to just support TIMESTAMP types; with a TODO comment mentioning the need for a `impala4` dialect.

mrdavidlaing: See 
dfe9d00

csringhofer (approved):

PR #445, opened 2021-05-13
Hack to fix non-valid utf-8 stings handling in Python 3

csringhofer:
The Thrift lib used with Python 3 was changed from thriftpy2 to
Apache Thrift in https://github.com/cloudera/impyla/pull/439
which broke the handling of STRING columns that contain values that
are not valid utf-8, as the library always throws an exception in
this case.

The hack to fix it is to change TStringColumn's underlying Thrift type
to 'binary' instead of 'string' and do the conversion in Impyla code.

Impala patches Thrift to solve the same issue in impala-shell,
see THRIFT-5303. I prefer a different solution for now for 2 reasons:
- THRIFT-5303 is only released in Thrift in 0.14.0, and bumping
  from 0.11.0 to 0.14.0 needs more consideration.
- The way it solves the issue (replacing invalid chars) is good
  for impala-shell where the final output is always text, but its
  lossiness makes it unsuitable if we actually want to transfer
  binary data in STRING.

Testing:
- added tests with SELECTs that return non-valid uft8 strings
- ran the tests with Python 2/3 + Apache Thrift/thriftpy2
  + BinaryProtocol/BinaryProtocolAccelerated
- saw some performance regression (5-10%) in Python 3, but it seems
  negligible compared to the ~6x gain from switching to Apache Thrift

### DIFF
--- a/impala/_thrift_gen/TCLIService/ttypes.py
+++ b/impala/_thrift_gen/TCLIService/ttypes.py
@@ -2348,7 +2348,7 @@ class TStringColumn(object):
                     self.values = []
                     (_etype93, _size90) = iprot.readListBegin()
                     for _i94 in range(_size90):
-                        _elem95 = iprot.readString()
+                        _elem95 = iprot.readBinary()
                         self.values.append(_elem95)
                     iprot.readListEnd()
                 else:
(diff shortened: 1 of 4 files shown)


stiga-huang:
+1

The solution makes sense to me. I think the clean solution can only wait until Impala support BINARY type. Then we can clearly define whether the function return type is STRING or BINARY that allows non-UTF8 characters.

PR #448, opened 2021-05-17
Fix TIMESTAMP handling regression in #445

danielewood:
TIMESTAMP was omitted from `_convert_strings_to_unicode`, causing `_parse_timestamp` to receive a bytes-like object for any returned column with a timestamp type.

Before:
```
2021-05-17 02:00:06,858 DEBUG sqlalchemy.pool.impl.QueuePool Created new connection <impala.hiveserver2.HiveServer2Connection object at 0x7f912df6ea50>
2021-05-17 02:00:06,859 DEBUG sqlalchemy.pool.impl.QueuePool Connection <impala.hiveserver2.HiveServer2Connection object at 0x7f912df6ea50> checked out from pool
2021-05-17 02:00:07,174 INFO sqlalchemy.engine.base.Engine SELECT type, fired_at FROM MyCompany.Zendesk.Zendesk_Webhook LIMIT 2
2021-05-17 02:00:07,175 INFO sqlalchemy.engine.base.Engine {}
2021-05-17 02:00:11,630 INFO sqlalchemy.engine.base.Engine ROLLBACK
Traceback (most recent call last):
  File "superset/test1.py", line 8, in <module>
    print(engine.execute("SELECT * FROM MyCompany.Zendesk.Zendesk_Webhook LIMIT 2").fetchall())
  File "/usr/local/lib/python3.7/site-packages/sqlalchemy/engine/result.py", line 1289, in fetchall
    e, None, None, self.cursor, self.context
  File "/usr/local/lib/python3.7/site-packages/sqlalchemy/engine/base.py", line 1514, in _handle_dbapi_exception
    util.raise_(exc_info[1], with_traceback=exc_info[2])
  File "/usr/local/lib/python3.7/site-packages/sqlalchemy/util/compat.py", line 182, in raise_
    raise exception
  File "/usr/local/lib/python3.7/site-packages/sqlalchemy/engine/result.py", line 1284, in fetchall
    l = self.process_rows(self._fetchall_impl())
  File "/usr/local/lib/python3.7/site-packages/sqlalchemy/engine/result.py", line 1230, in _fetchall_impl
    return self.cursor.fetchall()
  File "/usr/local/lib/python3.7/site-packages/impala/hiveserver2.py", line 553, in fetchall
    elements = self._pop_from_buffer(self.buffersize)
  File "/usr/local/lib/python3.7/site-packages/impala/hiveserver2.py", line 619, in _pop_from_buffer
    self._ensure_buffer_is_filled()
[...]

PR #449, opened 2021-05-17
Add test for timestamp handling fix in #448

attilajeges:
The timestamp handling regression was introduced
in #445 and fixed in #448.

### DIFF
--- a/impala/tests/test_data_types.py
+++ b/impala/tests/test_data_types.py
@@ -69,6 +69,32 @@ def test_date_basic(cur, date_table):
     results = cur.fetchall()
     assert results == [(datetime.date(1, 1, 1),), (datetime.date(1999, 9, 9),)]
 
+
+@yield_fixture(scope='module')
+def timestamp_table(cur):
+    table_name = 'tmp_timestamp_table'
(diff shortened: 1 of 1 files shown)


impala/tests/test_data_types.py
```diff
@@ -69,6 +69,32 @@ def test_date_basic(cur, date_table):
...
+    ddl = """CREATE TABLE {0} (ts timestamp)""".format(table_name)
+    cur.execute(ddl)
+    try:
+        yield table_name
+    finally:
+        cur.execute("DROP TABLE {0}".format(table_name))
+
+
+@pytest.mark.connect
+def test_timestamp_basic(cur, timestamp_table):
```
csringhofer, line 85: Can you also test timestamps with sub-second component? E.g. something like 
select cast("2021-05-20 00:00:00.123 " as timestamp), 
          cast("2021-05-20 00:00:00.123456 " as timestamp)
          cast("2021-05-20 00:00:00.123456789 " as timestamp);

datetime.datetime should be able to store them with microseconds precision, so the last one should not return the exact value.

PR #452, opened 2021-05-20
Add impala4 sqlalchemy dialect for Impala >= 4.0

attilajeges:


impala/tests/test_sqlalchemy.py
```diff
@@ -52,3 +52,32 @@ def test_sqlalchemy_compilation():
...
+                    Column('col5', DATE),
+                    Column('col6', VARCHAR(10)),
+                    impala_partition_by='HASH PARTITIONS 16',
+                    impala_stored_as='KUDU',
+                    impala_table_properties={
+                        'kudu.table_name': 'my_kudu_table',
+                        'kudu.master_addresses': 'kudu-master.example.com:7051'
+                    })
+    observed = str(CreateTable(mytable, bind=engine))
+    expected = ('\nCREATE TABLE mytable (\n\tcol1 STRING, \n\tcol2 TINYINT, '
```
csringhofer, line 78: This  test seems misleading to me, as this CREATE TABLE ... STORED AS KUDU would be rejected as there is no primary key column, which is mandatory in Kudu. Can you rewrite it to a non-Kudu, e.g. Parquet table? It should use impala_partitioned_by instead of impala_partition_by

attilajeges: Done

impala/tests/test_sqlalchemy.py
```diff
@@ -39,16 +39,47 @@ def test_sqlalchemy_compilation():
                     Column('col4', DOUBLE),
                     Column('col5', DATE),
                     Column('col6', VARCHAR(10)),
-                    impala_partition_by='HASH PARTITIONS 16',
-                    impala_stored_as='KUDU',
+                    impala_partitioned_by='HASH PARTITIONS 16',
```
csringhofer, line 42: This is only valid in Kudu - it should be something like (part_col int)

PR #453, opened 2021-06-02
Added missing expect_more_rows argument

dhanababum:
When I implementing get_log in my project found an issue.  
https://github.com/cloudera/impyla/blob/v0.17.0/impala/hiveserver2.py#L1345

_wrap_results takes three positional arguments but passing two arguments.


attilajeges (approved):

attilajeges:
Thanks for fixing this!

packet23:
I can confirm the change from this PR fixes the issue for me. @attilajeges would it be possible to release as 0.17.1 ?

PR #461, opened 2021-07-21
This allows multiple insert

darklord1807:
This enabled multiple insert which is supported by Impala and increases insert performance tremendously for huge inserts (42 inserts per second vs 5000 inserts per second).

  allToBeInserted.to_sql('dbtable', engine, if_exists='append', index=False,chunksize=2000,**method='multi')**

Multi insert allows passing of multiple values in a single INSERT clause (like a batch) and increases performance of insert

### DIFF
--- a/impala/sqlalchemy.py
+++ b/impala/sqlalchemy.py
@@ -190,6 +190,7 @@ class ImpalaDialect(DefaultDialect):
     max_identifier_length = 128
     supports_sane_rowcount = False
     supports_sane_multi_rowcount = False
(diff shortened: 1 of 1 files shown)


attilajeges (commented):
Thanks, LGTM. I don't have much experience with sqlalchemy. Could you provide some information regarding how you tested this?

csringhofer:
Closing this as https://github.com/cloudera/impyla/pull/499 was merged and it contains this change.

PR #463, opened 2021-07-27
Add support for authentication via JWT token

### LINKED ISSUE #462: Add support for authentication via JWT
Impala recently implemented a feature that allows for authentication via JWT over HTTP connections. See https://issues.apache.org/jira/browse/IMPALA-10489.
This allows a user to authenticate with a separate authentication service to obtain a JWT and pass that on communications with Impala.

Impyla should support passing the JWT in the Authentication Bearer HTTP header.

joemcdonnell:
Impala recently added support for authentication via a JWT
token in IMPALA-10489. This allows a user to authenticate
with a separate trusted authentication service, obtain
a JWT, and pass that JWT to Impala on the Authencation
Bearer HTTP header.

This implements support for connecting with auth_mechanism=JWT,
which requires a 'jwt' argument specifying the standard string
encoding for the JWT. This is passed on the Authentication Bearer
HTTP header on each message.

Fixes #462

Testing:
 - Added unit tests in test_dbapi_connect.py
 - Tested against a running Impala cluster with JWT authentication

### DIFF
--- a/impala/dbapi.py
+++ b/impala/dbapi.py
@@ -24,11 +24,12 @@ from impala.error import (  # noqa
     Error, Warning, InterfaceError, DatabaseError, InternalError,
     OperationalError, ProgrammingError, IntegrityError, DataError,
     NotSupportedError)
-from impala.util import warn_deprecate, warn_protocol_param
+from impala.util import (
+    warn_deprecate, warn_protocol_param, warn_unused_jwt, warn_nontls_jwt)
 import impala.hiveserver2 as hs2
 
 
-AUTH_MECHANISMS = ['NOSASL', 'PLAIN', 'GSSAPI', 'LDAP']
+AUTH_MECHANISMS = ['NOSASL', 'PLAIN', 'GSSAPI', 'LDAP', 'JWT']
 
 
 # PEP 249 module globals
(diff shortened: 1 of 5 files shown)


impala/dbapi.py
```diff
@@ -43,7 +44,7 @@ def connect(host='localhost', port=21050, database=None, timeout=None,
             ldap_user=None, ldap_password=None, use_kerberos=None,
             protocol=None, krb_host=None, use_http_transport=False,
             http_path='', auth_cookie_names=['impala.auth', 'hive.server2.auth'],
-            retries=3):
+            retries=3, jwt=None):
```
attilajeges, line 47: JWT tokens may also include expiration time in their payloads, so passing the JWT token directly to connect() may cause issues with long-running queries or when the same connection is used to execute several queries.

Maybe connect() could take a shell command instead that would return a fresh JWT token every time impyla makes an HTTP request?

(We do something similar for GSSAPI over HTTP, see transport.setGetCustomHeadersFunc() call in get_http_transport()).

attilajeges (commented):
Thanks for working on this.

impala/dbapi.py
joemcdonnell: I think there are going to be deployments that use JWTs where the expiration time is measured in hours or days. In that case, it is useful to have a simple option of just passing in the JWT. In particular, I think some clients won't know how to get a fresh JWT, because they assume that the JWT will last for a day and they are never running that long.

I agree that we will want a way to handle expiration more gracefully. It makes sense that a user could provide a callback that provides a JWT. For now, I think having the easy way that just passes in a JWT in a good place to start. We can always add additional parameters to connect() to provide different ways to maintain the JWT.

attilajeges: OK, thanks for the context. LGTM.

attilajeges (approved):

joemcdonnell:
Thanks for the review!

PR #482, opened 2021-10-16
Use assertTrue instead of failUnless for Python 3.11 compatibility.

tirkarthi:
The deprecated aliases were removed in Python 3.11 in python/cpython#28268 . `assertTrue` is present in both Python 2 and 3.


csringhofer:
Thanks for fixing this!

PR #494, opened 2022-07-13
Fix parameter substitution in execetutemany()

csringhofer:
Before this change the first execute_async called by executemany()
removed "paramstyle" from the configuration map, leading to applying
default (==all) substitution rules for subsequent parameters.

Testing:
- added a regression test (executemany() was totally untested)


wzhou-code:
Looks good to me

PR #496, opened 2022-08-29
Fix SSL connections (broken by THRIFT-5595)

csringhofer:


MikaelSmith (approved):

MikaelSmith:
Any way we could add a test around this?

csringhofer:
Added a commit with some basic tests.
Authentication tests are not completely automated in Impyla, as first an Impala cluster has to be set up.

I was thinking about how to automate these better - Impyla auth tests could be added to the Impala test suite, where running of custom clusters is convenient, or Impyla tests could be extended to restart the Impala cluster when needed. For the current fix I wouldn't dive into this deeper.

MikaelSmith (approved):

PR #498, opened 2022-09-09
enable a way to pass custom user_agent when using HTTP transport

tovganesh:
When using HTTP transport currently the Impyla driver hard codes the user_agent string as 'Python/ImpalaHttpClient'
This PR is for a way to configure this as a parameter from dbapi.connect :
<pre>
custom_user_agent = 'dbt/impala'
handle = impala.dbapi.connect(
                    host=credentials.host,
                    port=credentials.port,
                    auth_mechanism='LDAP',
                    use_http_transport=credentials.use_http_transport,
                    user=credentials.username,
                    password=credentials.password,
                    use_ssl=credentials.use_ssl,
                    http_path=credentials.http_path,
                    retries=credentials.retries,
                    user_agent=custom_user_agent
                )
</pre>
To avoid confusion with impala_shell, the default user_agent string is changed to 'Python/ImpylaHttpClient'

Testcase:
 impala/tests/test_dbapi_connect.py

After editing the warehouse parameters, run using:
pytest impala/tests/test_dbapi_connect.py -k "ImpalaConnectionTests and test_impala_ldap_connect_user_agent"
==================================================================================== test session starts =====================================================================================
platform darwin -- Python 3.9.12, pytest-7.1.3, pluggy-1.0.0
rootdir: /Users/ganesh.venkateshwara/code/cloudera/impyla
collected 11 items / 10 deselected / 1 selected

impala/tests/test_dbapi_connect.py .                                                                                                                                                   [100%]
======================================================================== 1 passed, 10 deselected, 1 warning in 11.02s ========================================================================

The test case merely check if the user_agent is accepted when connecting to impyla using ldap.

### DIFF
--- a/impala/_thrift_api.py
+++ b/impala/_thrift_api.py
@@ -151,6 +151,8 @@ class ImpalaHttpClient(TTransportBase):
     # new request.
     self.__custom_headers = None
     self.__get_custom_headers_func = None
+    # the default user agent if none is provied
+    self.__custom_user_agent = 'Python/ImpylaHttpClient'
 
   @staticmethod
   def basic_proxy_auth_header(proxy):
@@ -197,6 +199,9 @@ class ImpalaHttpClient(TTransportBase):
 
   def setCustomHeaders(self, headers):
     self.__custom_headers = headers
+
+  def setCustomUserAgent(self, user_agent):
+    self.__custom_user_agent = user_agent
 
   # Set callback function which generate HTTP headers for a specific auth mechanism.
   def setGetCustomHeadersFunc(self, func):
@@ -295,7 +300,7 @@ class ImpalaHttpClient(TTransportBase):
 
       self.refreshCustomHeaders()
       if not self.__custom_headers or 'User-Agent' not in self.__custom_headers:
-        user_agent = 'Python/ImpalaHttpClient'
+        user_agent = self.__custom_user_agent
         script = os.path.basename(sys.argv[0])
         if script:
           user_agent = '%s (%s)' % (user_agent, urllib.parse.quote(script))
@@ -368,7 +373,7 @@ def get_socket(host, port, use_ssl, ca_cert):
 def get_http_transport(host, port, http_path, timeout=None, use_ssl=False,
                        ca_cert=None, auth_mechanism='NOSASL', user=None,
                        password=None, kerberos_host=None, kerberos_service_name=None,
-                       http_cookie_names=None, jwt=None):
+                       http_cookie_names=None, jwt=None, user_agent=None):
     # TODO: support timeout
     if timeout is not None:
         log.error('get_http_transport does not support a timeout')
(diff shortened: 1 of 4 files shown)


impala/dbapi.py
```diff
@@ -101,6 +101,8 @@ def connect(host='localhost', port=21050, database=None, timeout=None,
         Currently cookie retention is supported for GSSAPI/LDAP/SASL/NOSASL/JWT over http.
     jwt: string containing a JSON Web Token
         This is used for auth_mechanism=JWT when using the HTTP transport.
+    user_agent: A user specified user agent when HTTP transport is used. If none is specified,
+        'Python/ImpalaHttpClient' is used
```
bartash, line 105: At the moment  'Python/ImpalaHttpClient' is used as the user agent for both impala shell and impyla. This was caused by code copying (maybe by me), but I think it's wrong, they should be different.  I think documenting that the user agent is 'Python/ImpalaHttpClient' would then make changing the default to be a breaking change, if we are being pedantic. So, please can you EITHER not document the default value here, OR change the default value to be something different (perhaps 'Python/ImpylaHttpClient').

bartash (commented):
Looks good, I have a request for a small change

impala/dbapi.py
tovganesh: Thanks for reviewing @bartash 
I have updated the default user agent to 'Python/ImpylaHttpClient' as per suggestion, which I felt was more correct thing to do.

bartash:
This looks good, can you squash your commits into a single commit please.

tovganesh:
> This looks good, can you squash your commits into a single commit please.

Please check if it commit looks good now.

impala/_thrift_api.py
```diff
@@ -151,6 +151,8 @@ def __init__(self, uri_or_host, port=None, path=None, cafile=None, cert_file=Non
     # new request.
     self.__custom_headers = None
     self.__get_custom_headers_func = None
+    # the default user agent if none is provied
+    self.custom_user_agent = 'Python/ImpylaHttpClient'
```
wzhou-code, line 155: define member variable as self.__custom_user_agent

impala/_thrift_api.py
```diff
@@ -295,7 +297,7 @@ def sendRequestRecvResp(data):
 
       self.refreshCustomHeaders()
       if not self.__custom_headers or 'User-Agent' not in self.__custom_headers:
-        user_agent = 'Python/ImpalaHttpClient'
+        user_agent = self.custom_user_agent
```
wzhou-code, line 300: use member variable as self.__custom_user_agent

impala/_thrift_api.py
```diff
@@ -368,7 +370,8 @@ def isOpen(self):
 def get_http_transport(host, port, http_path, timeout=None, use_ssl=False,
                        ca_cert=None, auth_mechanism='NOSASL', user=None,
                        password=None, kerberos_host=None, kerberos_service_name=None,
-                       http_cookie_names=None, jwt=None):
+                       http_cookie_names=None, jwt=None, user_agent=None):
+
```
wzhou-code, line 374: don't need extra line

impala/_thrift_api.py
tovganesh: done

wzhou-code:
Could you add an unit test case in tests/test_dbapi_connect.py ?

wzhou-code:
Thanks to work on this. It make sense to me.

PR #504, opened 2022-10-18
Use winkerberos on windows

### LINKED ISSUE #466: Windows Kerberos Support
Right now, using impyla on Windows with kerberos is difficult. If you're running Python v3.6 or v3.7 you can go to the infamous [Unofficial Windows Binaries for Python Extension Packages](https://www.lfd.uci.edu/~gohlke/pythonlibs/#sasl) and install the binary version of the sasl package and somewhat successfully install an older version of impyla with kerberos support. 

However, if you're forced to use a more recent version of Python say 3.8, you seem to be out of luck. 

It seems like impyla could follow the popular pattern of using the [winkerberos ](https://github.com/mongodb/winkerberos) package based on the platform detected to mitigate the issue. 

Would it be possible in the future to add support for this? Using CDH in a corporate environment is pretty painful without this feature.

jcrist:
This switches impyla to depend on `winkerberos` on windows rather than `kerberos`. The `winkerberos` package provides pre-built wheels for `windows` with the same python-facing API.

Fixes #466.

### DIFF
--- a/impala/_thrift_api.py
+++ b/impala/_thrift_api.py
@@ -436,7 +436,11 @@ def get_http_transport(host, port, http_path, timeout=None, use_ssl=False,
     elif auth_mechanism == 'GSSAPI':
         # For GSSAPI over http we need to dynamically generate custom request headers.
         def get_custom_headers(cookie_header, has_auth_cookie):
-            import kerberos
+            try:
+                import winkerberos as kerberos
+            except ImportError:
+                import kerberos
+
             custom_headers = {}
             if cookie_header:
                 log.debug('add cookies to HTTP header')
--- a/setup.py
+++ b/setup.py
@@ -43,7 +43,10 @@ setup(
     package_data={'impala.thrift': ['*.thrift']},
     install_requires=['six', 'bitarray', 'thrift==0.16.0', 'thrift_sasl==0.4.3'],
     extras_require={
-        "kerberos": ['kerberos>=1.3.0'],
+        "kerberos": [
+            'kerberos>=1.3.0;platform_system!="Windows"',
+            'winkerberos;platform_system=="Windows"',
+        ],
     },
     keywords=('cloudera impala python hadoop sql hdfs mpp spark pydata '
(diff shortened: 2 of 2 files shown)


csringhofer:
Thanks for working on this!
I agree that it would be much better to use a Kerberos module with prebuilt binaries.

I have concerns about testing though: I don't know whether anyone tests Impyla with Windows + Kerberos, so it seems possible to break some use cases without noticing it. All the test environments I know about are Linux based.
Are there relevant tests in the Ibis project?

jcrist:
>  Are there relevant tests in the Ibis project?

There are not, and I also have no way to test this. I've had similar code in place for years in another kerberos-related project though with no complaints, but this may just mean that no one is using it on windows :).

This PR was quick to work up, no pressure at all to merge it. Apologies that I can't be more help here ensuring this is correct.

csringhofer:
I started asking around about Impyla + Windows + Kerberos inside Cloudera.

I was thinking about an alternative in case switching to winkerberos from kerberos would be problematic for some people:
- the order of importing could reversed, so impyla would try winkerberos only if importing kerberos has failed
- a new extra name could be used, e.g. kerberos_with_winkerberos_on_windows or kerberos_prebuilt

This way an easy installation could be provided on Windows without the chance of breaking any existing use cases.

csringhofer:
>I started asking around about Impyla + Windows + Kerberos inside Cloudera.
I didn't get any feedback yet about this setup.
I am still concerned about possible breaking some existing workflows, so I would prefer  to use the solution I wrote above, so first try importing kerberos and then winkerberos (if os.name == 'nt').

@jcrist  Is it ok for you to proceed this way?

juliuscrn:
any change to have this merged? I can't use Ibis at work because kerberos needs to be built on Windows.

MinadorCH:
@csringhofer @jcrist Hi there! We are in need of this Patch for our internal setup (which only works with winkerberos). Can we help you with testing?

Issue #349, opened 2019-04-30
Numeric paramstyle substitution bug


MarechJ:
When using a list for passing parameters and one of the values contains a `:\d+` the value itself gets substituted. The bug applies to all paramstyles. See ipython logs below. I fixed it in: #348

This is the same as #317 

```
In [5]: cursor = conn.cursor()

In [6]: cursor.execute("INSERT INTO test_impyla VALUES (?, ?)", ['this will go wrong:2', 'im going to creep in'])
---------------------------------------------------------------------------
HiveServer2Error                          Traceback (most recent call last)
<ipython-input-6-4361277cbece> in <module>
----> 1 cursor.execute("INSERT INTO test_impyla VALUES (?, ?)", ['this will go wrong:2', 'im going to creep in'])

~/.pyenv/versions/3.6.5/envs/market-surveillance/lib/python3.6/site-packages/impala/hiveserver2.py in execute(self, operation, parameters, configuration)
    300         # PEP 249
    301         self.execute_async(operation, parameters=parameters,
--> 302                            configuration=configuration)
    303         log.debug('Waiting for query to finish')
    304         self._wait_to_finish()  # make execute synchronous

~/.pyenv/versions/3.6.5/envs/market-surveillance/lib/python3.6/site-packages/impala/hiveserver2.py in execute_async(self, operation, parameters, configuration)
    341             self._last_operation = op
    342
--> 343         self._execute_async(op)
    344
    345     def _debug_log_state(self):

~/.pyenv/versions/3.6.5/envs/market-surveillance/lib/python3.6/site-packages/impala/hiveserver2.py in _execute_async(self, operation_fn)
    360         self._reset_state()
    361         self._debug_log_state()
--> 362         operation_fn()
    363         self._last_operation_active = True
    364         self._debug_log_state()

~/.pyenv/versions/3.6.5/envs/market-surveillance/lib/python3.6/site-packages/impala/hiveserver2.py in op()
    338             op = self.session.execute(self._last_operation_string,
    339                                       configuration,
--> 340                                       async=True)
    341             self._last_operation = op
    342

~/.pyenv/versions/3.6.5/envs/market-surveillance/lib/python3.6/site-packages/impala/hiveserver2.py in execute(self, statement, configuration, async)
   1025                                    confOverlay=configuration,
   1026                                    runAsync=async)
-> 1027         return self._operation('ExecuteStatement', req)
   1028
   1029     def get_databases(self, schema='.*'):

[...]

PR #508, opened 2023-01-09
Fixed numeric parameter substitution bug

gjask:
I attempted to finally solve problem with multiple params substitution when one pass of substitution unexpectedly creates new markers as specified in https://github.com/cloudera/impyla/issues/349

I decided on doing single pass for all of positional/numeric paramstyles when no particular paramstyle is passed in configuration. This closely follows previous behaviour without interpreting "new markers" created by previous substitutions.


gjask:
I have tested `test_query_parameters.py` on py3.11 and py2.7

PR #511, opened 2023-05-22
Add get_view_name support to SQLAlchemy

huw0:
This fixes SQLAlchemy apps that expect `get_view_names` to be implemented.

Since `SHOW TABLES` doesn't distinguish between tables and views in Impala, this will always return an empty list.

Fixes apache/superset#23850

### DIFF
--- a/impala/sqlalchemy.py
+++ b/impala/sqlalchemy.py
@@ -244,6 +244,11 @@ class ImpalaDialect(DefaultDialect):
             for tup in connection.execute(query).fetchall()
         ]
         return tables
+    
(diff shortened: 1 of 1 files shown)


giuturtu:
It works ! Used to fix superset

Heyzi (approved):
works fine!

wzhou-code (approved):
Looks good to me

uniqueinput:
Its worth to mention that SHOW VIEWS is expected to be supported in Impala soon
https://gerrit.cloudera.org/#/c/20669/

PR #513, opened 2023-05-26
Add cursor configuration to connection args

VitaliyFedorov:
It's feature for possibity to add cursor configuration options while connection creating, to be used for any new cursor further.

Reason: Some options much better to set once, but not each time when you need cursor, e.g.: `impala.doas.user`, which allow to set proxy user for impala (https://impala.apache.org/docs/build/html/topics/impala_delegation.html)

### DIFF
--- a/impala/hiveserver2.py
+++ b/impala/hiveserver2.py
@@ -127,8 +127,11 @@ class HiveServer2Connection(Connection):
 
         log.debug('.cursor(): getting new session_handle')
 
-        session_configuration = {**self.cursor_configuration, **configuration}
-        session = self.service.open_session(user, session_configuration)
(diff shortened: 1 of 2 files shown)


impala/hiveserver2.py
```diff
@@ -126,7 +127,11 @@ def cursor(self, user=None, configuration=None, convert_types=True,
 
         log.debug('.cursor(): getting new session_handle')
 
-        session = self.service.open_session(user, configuration)
+        session_configuration = {}
```
csringhofer, line 130: The comments about configuration could be updated to reflect that it overrides connection level cursor_configuration

csringhofer (changes requested):

PR #515, opened 2023-06-01
IMPALA-12149: Adds additional checks to ensure valid connection arguments

jasonmfehr:
IMPALA-12149: Adds additional checks to ensure connection arguments that are applicable to username/password auth and JWT auth are not mixed together on the same call to the connect method.

These additional checks prevent confusion about which authentication method is actually used for the connection.

New tests were added to cover the new checks.

### DIFF
--- a/impala/tests/test_dbapi_connect.py
+++ b/impala/tests/test_dbapi_connect.py
@@ -25,7 +25,7 @@ else:
 
 import pytest
 
-from impala.dbapi import connect
+from impala.dbapi import connect, AUTH_MECHANISMS
 from impala.util import _random_id
 from impala.tests.util import ImpylaTestEnv, SocketTracker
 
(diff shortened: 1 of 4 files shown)


jasonmfehr:
Ran the `params_neg` tests in `impala/tests/test_dbapi_connect.py` on Python 2.7 and 3.8.  Tests passed on both versions of Python.

impala/tests/test_dbapi_connect.py
```diff
@@ -166,24 +166,74 @@ def test_jwt_auth(self):
...
+        def run_test(auth_mechanism):
+            try:
+                connect(auth_mechanism=auth_mechanism, jwt="dummy.jwt.arg")
+                assert False, "'connect' method should have thrown an exception but did not"
+            except NotSupportedError as e:
+                assert "'jwt' argument cannot be specified with '{0}' authentication".format(auth_mechanism) in str(e)
+
+        for auth_mech in AUTH_MECHANISMS:
+            if auth_mech is not "JWT":
+              run_test(auth_mech)
```
csringhofer, line 199: nit: inconsistent indentation

csringhofer (approved):

PR #516, opened 2023-06-04
Support CHAR type in SQLAlchemy

huw0:
This adds CHAR support which fixes errors in tables or views with CHAR type when using Impyla with Apache Superset.


csringhofer (approved):
Thanks for working on this!

Issue #518, opened 2023-07-19
Backport changes from [IMPALA-12294] ("Cookie handling broken In Impala Shell with python 3") to Impyla


bartash:
I observer that the cookie code is the same as in Impala Shell so almost certainly we need this change

bartash:
Fixed in upstream

PR #519, opened 2023-08-02
Issue #518: Fix Cookie handling with Python 3 [from IMPALA-12294]

bartash:
Impyla gets cookies from an HTTMessage object formed from a response to an HTTP message. The format of cookies in the message differs across the python versions. In Python 2 the HTTPMessage is a mimetools.Message object, and the Set-Cookie values all appear in a single header, separated by newlines. In Python 3 the HTTPMessage is an email.message.Message, and the Set-Cookie values appear as duplicate headers.

Add platform dependent code to get_all_matching_cookies() that loads cookies from all the Set-Cookie headers.

TESTING:
    Changed test_get_all_matching_cookies() to build the HTTPMessage
    using a new utility method that creates Set-Cookie headers in
    the appropriate format for the platform.

    I hand tested with a proxy that inserted 3 cookies into http
    responses. I added the 3 cookie names to the list of default
    cookies. I ran TestHttpConnect.test_simple_connect() connecting
    to Impala through the proxy and verified with the debugger that
    the cookies were returned correctly from
    get_all_matching_cookies() in both python2 and python3.


impala/util.py
```diff
@@ -202,13 +203,19 @@ def get_cookie_expiry(c):
 
 
 def get_all_matching_cookies(cookie_names, path, resp_headers):
+    matching_cookies = None
```
wzhou-code, line 206: Why add this line

bartash: Thanks, it does nothing, removing in next patch

wzhou-code:
Looks good to me

wzhou-code (approved):

PR #521, opened 2023-10-19
[Docs] Add comment to help Hive users quick start

NicolasPA:
Current Usage part works well for Impala users but will fail for Hive users because of the `auth_mechanism` default value. 
This adds a comment targeted towards Hive users so they can quick start too.


csringhofer (approved):
Thanks for improving the docs!

PR #525, opened 2024-01-03
Add Knox cookies in default cookies list

cravani:
Connecting over Knox backed with multiple HiveServer2 Instances and Load-balancing enabled leads to Invalid Session Handle error.
The root cause of this issue is because Knox uses different cookies than default ones i.e impala.auth, impala.session.id, hive.server2.auth

This can easily be fixed by adding below property in connection string.
`http_cookie_names = ["KNOXSESSIONID","KNOX_BACKEND-HIVE"]`

It would be good to have them added to default list and avoid inconvenience.

### DIFF
--- a/impala/dbapi.py
+++ b/impala/dbapi.py
@@ -191,8 +191,8 @@ def connect(host='localhost', port=21050, database=None, timeout=None,
         warn_deprecate('auth_cookie_names', 'http_cookie_names')
         http_cookie_names = auth_cookie_names
     elif http_cookie_names is None:
-        # Set default value as the list of HTTP cookie names used by Impala and Hive.
-        http_cookie_names = ['impala.auth', 'impala.session.id', 'hive.server2.auth']
+        # Set default value as the list of HTTP cookie names used by Impala, Hive and Knox.
+        http_cookie_names = ['impala.auth', 'impala.session.id', 'hive.server2.auth', 'KNOXSESSIONID', 'KNOX_BACKEND-HIVE', 'JSESSIONID']
 
     service = hs2.connect(host=host, port=port,
                           timeout=timeout, use_ssl=use_ssl,


impala/dbapi.py
```diff
@@ -192,7 +192,7 @@ def connect(host='localhost', port=21050, database=None, timeout=None,
         http_cookie_names = auth_cookie_names
     elif http_cookie_names is None:
         # Set default value as the list of HTTP cookie names used by Impala and Hive.
```
wzhou-code, line 194: Could you update the comments for three new Knox cookies? Are they used for session management? For cookie-based authentication, cookie name must end with '.auth'.

impala/dbapi.py
```diff
@@ -192,7 +192,7 @@ def connect(host='localhost', port=21050, database=None, timeout=None,
...
-        http_cookie_names = ['impala.auth', 'impala.session.id', 'hive.server2.auth']
+        http_cookie_names = ['KNOXSESSIONID', 'KNOX_BACKEND-HIVE', 'JSESSIONID', 'impala.auth', 'impala.session.id', 'hive.server2.auth']
```
wzhou-code, line 195: Move the three new cookie names to the end of array.

impala/dbapi.py
cravani: @wzhou-code  Thank you for the review, I have updated the comments and moved them to the end. Those cookies are indeed used by Knox and are part of defaulg cookies list for ODBC Driver too.
Ref [Page 98]: https://docs.cloudera.com/documentation/other/connectors/hive-odbc/2-7-0/Cloudera-ODBC-Driver-for-Apache-Hive-Install-Guide.pdf

Screenshot 2024-01-03 at 1 08 30 PM

impala/dbapi.py
cravani: Moved the new cookie names to the end of array.

wzhou-code (approved):

wzhou-code:
looks good

PR #530, opened 2024-01-30
Add tox.ini to help testing with multiple python versions

csringhofer:


csringhofer:
It turned out that the current version doesn't actually test with Python 2.7, for some reason it doesn't detect it on my system.
For this reason I didn't catch that the versioneer bump breaks python 2.7: https://github.com/cloudera/impyla/pull/532

csringhofer:
Update the PR to support Python 2.7 too. This needed setting virtualenv<20.22.0, as 20.22.0 dropped python 2 support.

rizaon (approved):
Ran `tox` in my local machine and all pass. LGTM.

Issue #529, opened 2024-01-23
Error in Python 3.12.1 trying to run connect


jmagana2000:
Got this exception when trying to create a connect object when running in Docker container using Ubuntu 22.04 and Python 3.12.1

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.12/site-packages/impala/dbapi.py", line 197, in connect
    service = hs2.connect(host=host, port=port,
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/impala/hiveserver2.py", line 865, in connect
    transport.open()
  File "/usr/lib/python3.12/site-packages/thrift/transport/TTransport.py", line 155, in open
    return self.__trans.open()
           ^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/impala/_thrift_api.py", line 174, in open
    self.__http = http_client.HTTPSConnection(self.host, self.port,
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: HTTPSConnection.__init__() got an unexpected keyword argument 'key_file'

csringhofer:
Merged https://github.com/cloudera/impyla/pull/531 so this should be solved.

Issue #362, opened 2019-08-14
Add HTTP server authentication when moving to Thrift 0.12


lekv:
#359 will add endpoint support but server authentication is only available in Thrift 0.12. We should support it, once we depend on Thrift 0.12.

danjampro:
Hi @lekv, are there any updates on this?

lekv:
No updates, I haven't been working on Impala for a while. Your best bet would be to ask on one of the mailing lists or on Slack. See https://impala.apache.org/community.html for the details.

csringhofer:
A fix was merged and released in 0.23a1

lekv:
That's awesome, thank you for updating this Csaba!

PR #531, opened 2024-01-30
Issue #529: Fix https connection with Python 3.12

csringhofer:
Python 3.12 removed deprecated certfile and key_file arguments from http_client.HTTPSConnection. These should be always empty in Impyla as the server is never verified in https connections (see #362).

### DIFF
--- a/impala/_thrift_api.py
+++ b/impala/_thrift_api.py
@@ -171,9 +171,15 @@ class ImpalaHttpClient(TTransportBase):
       self.__http = http_client.HTTPConnection(self.host, self.port,
                                                timeout=self.__timeout)
     elif self.scheme == 'https':
+      # (#529) From python 3.12 http_client.HTTPSConnection no longer
+      # expects certfile and key_file. Certfile is included in context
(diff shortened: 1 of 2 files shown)


rizaon (approved):
Tested with python 2.7, 3.8, and 3.12.
test_hs2_fault_injection.py show warnings about pytest.PytestRemovedIn8Warning with python 3, but unrelated with this patch.

csringhofer:
>test_hs2_fault_injection.py show warnings about pytest.PytestRemovedIn8Warning with python 3, but unrelated with this patch.
Did you use my tox patch (#530) to run the tests? If not, what was the pytest version?
I also saw a lot of deprecation warnings when running with newer pytest version.
It would be nice to fix deprecations and probably switch to a newer version of pytest (the tox patch uses 6.*, while Impala uses 2.9.2). This may not be possible if we want to support Python versions that are not compatible with newer pytest versions,

rizaon:
I did not, so pip was pulling pytest 8. I tried again with pytest 6 and all pass without warnings.

PR #532, opened 2024-01-30
Update versioneer to 0.29 (needed for Python 3.12)

csringhofer:
The old version used deprecated functions that were removed in Python 3.12.

The change only contains code generated by:
versioneer install


rizaon (approved):
I did `versioneer install` myself and the change set match with this.

csringhofer:
> I did `versioneer install` myself and the change set match with this.

Thanks for checking!

PR #533, opened 2024-02-02
Revert versioneer bump

csringhofer:


csringhofer:
Merging this quickly as it broke some use cases where pip was using master version of Impyla.

pp-gborodin:
This also breaks python 3.12 support, any updates on this?

csringhofer:
The plan is to:
1. release a final impyla version that supports Python 2.7
2. reapply the patch to make it work with Python 3.12
3. create a release that supports Python 3.12

I am working on 1, but spotted a bug during testing (#549), waiting for someone to review the fix.

PR #535, opened 2024-02-14
Turn regex strings into raw strings

hackermandh:
I updated some minor versions of some dependencies in my own application and the following error popped up. I checked other strings that were used with the `re` module, and those were all raw strings, so I checked and if I manually (and locally) changed the string to a raw string, everything ran fine.

```
E     File "/home/david/workon/impala-39/lib/python3.9/site-packages/impala/sqlalchemy.py", line 227
E       m = re.match('.*?(\d{1,3})\.(\d{1,3})\.(\d{1,3}).*', v)
E                    ^
E   SyntaxError: invalid escape sequence \d
```

Not sure why updating dependencies broke this code specifically.

### DIFF
--- a/impala/interface.py
+++ b/impala/interface.py
@@ -270,7 +270,7 @@ def _bind_parameters_dict(operation, parameters):
             string_parameters[name] = str(value)
 
     # replace named parameters by their pyformat equivalents
-    operation = re.sub(":([^\d\W]\w*)", "%(\g<1>)s", operation)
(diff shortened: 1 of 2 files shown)


hackermandh:
@csringhofer I saw a pre-version was just being released - if this tiny fix could be included in the next version, that would be amazing :)

csringhofer:
Thanks for the fix!

PR #538, opened 2024-04-09
Fix SQLAlchemy support for Impala on Python 3.10

jvprosser:
1. Add text() wrapper for metadata queries.
2. Remove table names from get_column output


jvprosser:
re-added 'r'

csringhofer:
Thanks for fixing this, the change looks good!

PR #545, opened 2024-08-06
Log fake password instead of actual password to avoid leaking sensitive info

WWakker:
At the moment the actual password is logged at debug level. It is not good practice to log sensitive information.


csringhofer (approved):
Ouch, thanks for fixing this

Issue #543, opened 2024-07-18
Allow skipping utf8 converison in Python3


csringhofer:
Ran some benchmarks on local Impala dev environment:
Query: select * from tpch_parquet.lineitem"
returns 6 million rows (numbers, strings)

ClientFetchWaitTimer looks like this:
Impyla (Python2, fetchcolumnar())                 7s967ms
Impyla (Python3, fetchcolumnar()):              14s163ms
Impyla (Python2, fetchmany(1024)):            17s791ms
Impyla (Python3, fetchmany(1024)):            23s816ms

Based on cProfile the Python2 vs 3 differences come from _convert_strings_to_unicode().

Adding a cursor option like "convert_strings_to_unicode" similarly to "convert_types" would be useful.

PR #548, opened 2024-08-12
Allow skipping utf8 converison in Python3

mszjat:
https://github.com/cloudera/impyla/issues/543

### DIFF
--- a/impala/hiveserver2.py
+++ b/impala/hiveserver2.py
@@ -1055,11 +1055,12 @@ class CBatch(Batch):
             # STRING columns are read as binary and decoded here to be able to handle
             # non-valid utf-8 strings in Python 3.
 
-            if convert_strings_to_unicode:
-                self._convert_strings_to_unicode(type_, is_null, values, 
-                    types=["STRING", "LIST", "MAP", "STRUCT", "UNIONTYPE", "NULL", "VARCHAR", "CHAR", "TIMESTAMP", "DECIMAL", "DATE"])
-            elif convert_types and six.PY3:
-                self._convert_strings_to_unicode(type_, is_null, values, types=["TIMESTAMP", "DECIMAL", "DATE"])
+            if six.PY3:
+                if convert_strings_to_unicode:
+                    self._convert_strings_to_unicode(type_, is_null, values, 
+                        types=["STRING", "LIST", "MAP", "STRUCT", "UNIONTYPE", "NULL", "VARCHAR", "CHAR", "TIMESTAMP", "DECIMAL", "DATE"])
+                elif convert_types
+                    self._convert_strings_to_unicode(type_, is_null, values, types=["TIMESTAMP", "DECIMAL", "DATE"])
 
             if convert_types:
                 values = self._convert_values(type_, is_null, values)
--- a/impala/hiveserver2.py
+++ b/impala/hiveserver2.py
@@ -84,7 +84,8 @@ class HiveServer2Connection(Connection):
         raise NotSupportedError
 
     def cursor(self, user=None, configuration=None, convert_types=True,
-               dictify=False, fetch_error=True, close_finished_queries=True):
+               convert_strings_to_unicode=False, dictify=False, 
+               fetch_error=True, close_finished_queries=True):
         """Get a cursor from the HiveServer2 (HS2) connection.
 
         Parameters
@@ -96,6 +97,9 @@ class HiveServer2Connection(Connection):
             When `False`, timestamps and decimal values will not be converted
             to Python `datetime` and `Decimal` values. (These conversions are
             expensive.) Only applies when using HS2 protocol versions > 6.
+        convert_strings_to_unicode : bool, optional
+            When `False`, string values will not be converted to unicode. 
+            Conversion may take significant time on large datasets.
         dictify : bool, optional
             When `True` cursor will return key value pairs instead of rows.
         fetch_error : bool, optional
@@ -150,6 +154,7 @@ class HiveServer2Connection(Connection):
         cursor_class = HiveServer2DictCursor if dictify else HiveServer2Cursor
 
         cursor = cursor_class(session, convert_types=convert_types,
+                              convert_strings_to_unicode=convert_strings_to_unicode,
                               fetch_error=fetch_error,
                               close_finished_queries=close_finished_queries)
 
@@ -168,9 +173,10 @@ class HiveServer2Cursor(Cursor):
     # HiveServer2Cursor objects are associated with a Session
     # they are instantiated with alive session_handles
 
-    def __init__(self, session, convert_types=True, fetch_error=True, close_finished_queries=True):
+    def __init__(self, session, convert_types=True, convert_strings_to_unicode=False, fetch_error=True, close_finished_queries=True):
         self.session = session
         self.convert_types = convert_types
+        self.convert_strings_to_unicode = convert_strings_to_unicode
         self.fetch_error = fetch_error
         self.close_finished_queries = close_finished_queries
 
@@ -570,7 +576,8 @@ class HiveServer2Cursor(Cursor):
             batch = (self._last_operation.fetch(
                          self.description,
                          self.buffersize,
-                         convert_types=self.convert_types))
+                         convert_types=self.convert_types,
+                         convert_strings_to_unicode=self.convert_strings_to_unicode))
             if len(batch) == 0:
                return None
             return batch
@@ -620,7 +627,8 @@ class HiveServer2Cursor(Cursor):
             batch = (self._last_operation.fetch(
                          self.description,
                          self.buffersize,
-                         convert_types=self.convert_types))
+                         convert_types=self.convert_types,
+                         convert_strings_to_unicode=self.convert_strings_to_unicode))
             if len(batch) == 0:
                 break
             batches.append(batch)
@@ -660,7 +668,8 @@ class HiveServer2Cursor(Cursor):
                           '=> fetching more data')
                 self._buffer = self._last_operation.fetch(self.description,
                                                           self.buffersize,
-                                                          convert_types=self.convert_types)
+                                                          convert_types=self.convert_types,
+                                                          convert_strings_to_unicode=self.convert_strings_to_unicode)
                 if len(self._buffer) > 0:
                     return
                 if not self._buffer.expect_more_rows:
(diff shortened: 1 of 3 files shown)


impala/hiveserver2.py
```diff
@@ -84,7 +84,8 @@ def rollback(self):
         raise NotSupportedError
 
     def cursor(self, user=None, configuration=None, convert_types=True,
-               dictify=False, fetch_error=True, close_finished_queries=True):
+               convert_strings_to_unicode=False, dictify=False,
```
csringhofer, line 87: To guarantee backward compatibility the new argument should be the last in the list + it should be True by default.

impala/hiveserver2.py
```diff
@@ -1012,7 +1021,7 @@ def pop_to_preallocated_list(self, output_list, count, offset=0, stride=1):
 
 class CBatch(Batch):
 
-    def __init__(self, trowset, expect_more_rows, schema, convert_types=True):
+    def __init__(self, trowset, expect_more_rows, schema, convert_types=True, convert_strings_to_unicode=False):
```
csringhofer, line 1024: Similarly to cursor(), this should be True by default.

impala/hiveserver2.py
```diff
@@ -1040,7 +1049,7 @@ def __init__(self, trowset, expect_more_rows, schema, convert_types=True):
 
             # STRING columns are read as binary and decoded here to be able to handle
             # non-valid utf-8 strings in Python 3.
-            if six.PY3:
+            if six.PY3 and convert_strings_to_unicode:
```
csringhofer, line 1052: The case when convert_strings_to_unicode is False but convert_types is True also needs to be consider - the code in _convert_values probably expects Unicode when converting DATE/TIMESTAMP/DECIMAL types.

impala/hiveserver2.py
```diff
@@ -96,6 +97,9 @@ def cursor(self, user=None, configuration=None, convert_types=True,
             When `False`, timestamps and decimal values will not be converted
             to Python `datetime` and `Decimal` values. (These conversions are
             expensive.) Only applies when using HS2 protocol versions > 6.
+        convert_strings_to_unicode : bool, optional
+            When `False`, string values will not be converted to unicode.
```
csringhofer, line 101: The return type could be mentioned both for the True/False case.
+ It should be mentioned that this has effect only in Python3.

csringhofer (commented):
Looks good in general, left a few comments.
It would be also nice to have some tests about getting the expected Python types for the different values of this parameter. There are already some similar tests in test_data_types.py

impala/hiveserver2.py
mszjat: Fixed

impala/hiveserver2.py
mszjat: Agree, fixed.

impala/hiveserver2.py
mszjat: Makes sense, please check out new design.

impala/tests/conftest.py
```diff
@@ -113,3 +113,9 @@ def cur(con):
     cur = con.cursor()
     yield cur
     cur.close()
+
+@fixture(scope='session')
+def cur_noconv(con):
+    cur = con.cursor(convert_types=True, convert_strings_to_unicode=False)
```
csringhofer, line 119: here and in test_data_types.py: is there a reason for removing the final new lines?

mszjat: It's by mistake, fixed

impala/hiveserver2.py
```diff
@@ -1040,14 +1054,33 @@ def __init__(self, trowset, expect_more_rows, schema, convert_types=True):
 
             # STRING columns are read as binary and decoded here to be able to handle
             # non-valid utf-8 strings in Python 3.
+
             if six.PY3:
-                self._convert_strings_to_unicode(type_, is_null, values)
+                if convert_strings_to_unicode:
+                    self._convert_strings_to_unicode(type_, is_null, values, 
+                        types=["STRING", "LIST", "MAP", "STRUCT", "UNIONTYPE", "NULL", "VARCHAR", "CHAR", "TIMESTAMP", "DECIMAL", "DATE"])
```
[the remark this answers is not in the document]
mszjat: Absolutely, fixed

impala/hiveserver2.py
```diff
@@ -96,6 +97,12 @@ def cursor(self, user=None, configuration=None, convert_types=True,
             When `False`, timestamps and decimal values will not be converted
             to Python `datetime` and `Decimal` values. (These conversions are
             expensive.) Only applies when using HS2 protocol versions > 6.
+        convert_strings_to_unicode : bool, optional
+            When `True`, the following types will be converted to unicode:
```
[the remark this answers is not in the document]
mszjat: Added.

impala/tests/conftest.py
```diff
@@ -113,3 +113,9 @@ def cur(con):
     cur = con.cursor()
     yield cur
     cur.close()
+
+@fixture(scope='session')
+def cur_noconv(con):
```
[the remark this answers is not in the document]
mszjat: Agree.

impala/tests/test_data_types.py
```diff
@@ -54,43 +54,47 @@ def date_table(cur):
...
+                   values (date "0001-01-01"), (date "1999-9-9")'''.format(table_name))
     try:
         yield table_name
     finally:
         cur.execute("DROP TABLE {0}".format(table_name))
 
-
-@pytest.mark.connect
-def test_date_basic(cur, date_table):
+def setup_test_date(cur, date_table):
```
csringhofer, line 64: The naming doesn't make sense to me - this function does a select and checks the result, so it is not a setup function.

impala/tests/test_data_types.py
```diff
@@ -61,41 +61,40 @@ def date_table(cur):
...
+                          (cast("9999-12-31 23:59:59" as timestamp))'''.format(table_name))
     try:
         yield table_name
     finally:
         cur.execute("DROP TABLE {0}".format(table_name))
 
 
-@pytest.mark.connect
-def test_timestamp_basic(cur, timestamp_table):
+def setup_test_timestamp(cur, timestamp_table):
```
csringhofer, line 96: Same as for dates, at this point this function is not about setup.

impala/tests/test_data_types.py
```diff
@@ -54,43 +54,47 @@ def date_table(cur):
...
     cur.execute('select d from {0} order by d'.format(date_table))
     results = cur.fetchall()
     assert results == [(datetime.date(1, 1, 1),), (datetime.date(1999, 9, 9),)]
 
+@pytest.mark.connect
+def test_date_basic(cur, date_table):
+    setup_test_date(cur, date_table)
+
+@pytest.mark.connect
+def test_date_no_string_conv(cur_no_string_conv, date_table):
```
csringhofer, line 75: Can you also add similar test for decimals?

impala/tests/test_data_types.py
```diff
@@ -120,3 +131,24 @@ def test_utf8_strings(cur):
     result = cur.fetchone()[0]
     assert result == b"\xaa"
     assert result.decode("UTF-8", "replace") == u"�"
+
+@pytest.mark.connect
+def test_utf8_strings_no_string_conv(cur_no_string_conv):
+    cur = cur_no_string_conv
+    """Use STRING/VARCHAR/CHAR values  with multi byte unicode code points in a query."""
```
csringhofer, line 138: I don't think that this complex unicode handling is needed for the "no conversion" case. The main point should be to check that with unicode conversion the type returned is unicode, while without it it is bytes (in Python 3)

mszjat: Got it

impala/tests/test_data_types.py
mszjat: Sure

impala/tests/test_data_types.py
mszjat: Renamed

impala/tests/test_data_types.py
```diff
@@ -49,6 +51,47 @@ def test_cursor_description_precision_scale(cur, decimal_table):
...
+        yield table_name
+    finally:
+        cur.execute("DROP TABLE {0}".format(table_name))
+
+
+def common_test_decimal(cur, decimal_table):
+    """Read back a few decimal values in a wide range."""
+    cur.execute('select val from {0} order by val'.format(decimal_table))
+    results = cur.fetchall()
+    assert results == [(Decimal('-999999999.999999999'),),
```
csringhofer, line 77: nice to have: it would increase the coverage of the tests if there would be also NULL values in these test tables

mszjat: Agree, added

impala/tests/test_data_types.py
```diff
@@ -132,23 +182,22 @@ def test_utf8_strings(cur):
...
-    assert result == (b"\xe5\xbc\x95\xe6",) * 3
-    assert result[0].decode("UTF-8", "replace") == u"引�"
 
-    cur.execute('select unhex("AA")')
-    result = cur.fetchone()[0]
-    assert result == b"\xaa"
-    assert result.decode("UTF-8", "replace") == u"�"
+
+    if sys.version_info[0] < 3:
+        assert result[0] == u"Test string"
```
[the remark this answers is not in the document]
mszjat: Fixed.

impala/tests/test_data_types.py
```diff
@@ -120,3 +183,22 @@ def test_utf8_strings(cur):
...
+
+
+@pytest.mark.connect
+def test_string_no_string_conv(cur_no_string_conv):
+    cur = cur_no_string_conv
+    cur.execute('select "Test string"')
+    result = cur.fetchone()
+
+    if sys.version_info[0] < 3:
+        assert not isinstance(result[0], unicode)
```
csringhofer, line 202: isinstance with str would be clearer

impala/tests/test_data_types.py
```diff
@@ -49,6 +51,47 @@ def test_cursor_description_precision_scale(cur, decimal_table):
     for (exp, obs) in zip(expected, observed):
         assert exp == obs
 
+
+@fixture(scope='module')
+def decimal_table2(cur):
```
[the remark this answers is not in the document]
csringhofer: ack

impala/tests/test_data_types.py
mszjat: fixed

Issue #549, opened 2024-08-13
Retry logic can incorrectly repeat execution in non-http connections


csringhofer:
Found this during the investigation of #547:
- the slow DROP TABLE in Impala led to hitting the 5 sec timeout in the test
  - note that the timeout was not caused by network, but simply the time taken by the synchronous execute_statement
- in reality 2 DROP TABLE statements were executed in Impala:
  - the 1. failing with timeout
  - the 2. failing with "table doesn't exist", as the 1. was actually executed by Impala, only the client timeouted

Was also able to reproduce it with INSERT statements, sometimes even executing it 3 times, so I consider this to be a very serious correctness issue.
Not sure if other (network) errors can also trigger this, or only timeouts.

The bug is in ThriftRpc._execute(): https://github.com/cloudera/impyla/blob/9a853c1e7d229342bdbe491fd4c5955daa5c4ebb/impala/hiveserver2.py#L1167
The intention seems to be to only retry connection failures, but exceptions from the actual RPC also lead to retry, which is not ok if the RPC is not idempotent, as coordinator may be already executing the statement at this point.

http connections only retry idempotent RPCs (#424), so the issue doesn't happen there.

The issue seems to be in Impala since a very long time - the problematic code was added in #153, but my impression is that the bug existed even before the refactor. It is possible that bumping Thrift versions also affected this by changing the type of exception thrown. Only checked with 0.20a1 and 0.19.0 so far.

PR #550, opened 2024-08-13
Avoid retrying non-idempotent RPCs in binary connections (#549)

csringhofer:
See the #549 for the detailed analyses of the issue.

The fix works similarly to the existing solution for http connections:
- each RPC knows whether it is idempotent
- if the error comes from establishing the connection, then retry
- if the error comes from executing the RPC, only retry if the RPC is idempotent

A test is added that relies on slow metadata handling in the Impala cluster to trigger timouts. It would be nice to add wider and more reliable tests in the future similarly to the http tests in test_hs2_fault_injection.py

### DIFF
--- a/impala/hiveserver2.py
+++ b/impala/hiveserver2.py
@@ -1167,12 +1167,7 @@ class ThriftRPC(object):
                 log.debug('Transport opened')
                 func = getattr(self.client, func_name)
                 return func(request)
-            except socket.error as e:
-                if open_finished and not safe_to_retry: raise e
-                msg = "RPC failed" if open_finished else "Failed to open transport"
-                log.exception('%s (tries_left=%s)', msg, tries_left)
-                last_exception = e
-            except TTransportException as e:
+            except (socket.error, TTransportException) as e:
                 if open_finished and not safe_to_retry: raise e
                 msg = "RPC failed" if open_finished else "Failed to open transport"
                 log.exception('%s (tries_left=%s)', msg, tries_left)
@@ -1210,7 +1205,7 @@ class ThriftRPC(object):
             open_finished = False
             tries_left -= 1
 
-        if last_exception is not None:
+        if last_exception:
             raise last_exception
         raise HiveServer2Error('Failed after retrying {0} times'
                                .format(self.retries))
--- a/impala/hiveserver2.py
+++ b/impala/hiveserver2.py
@@ -1144,41 +1144,45 @@ class ThriftRPC(object):
         self.client = client
         self.retries = retries
 
-    def _rpc(self, func_name, request, retry_on_http_error=False):
+    def _rpc(self, func_name, request, safe_to_retry=False):
         self._log_request(func_name, request)
-        response = self._execute(func_name, request, retry_on_http_error)
+        response = self._execute(func_name, request, safe_to_retry)
         self._log_response(func_name, response)
         err_if_rpc_not_ok(response)
         return response
 
-    def _execute(self, func_name, request, retry_on_http_error=False):
+    def _execute(self, func_name, request, safe_to_retry=False):
         # pylint: disable=protected-access
         # get the thrift transport
         transport = self.client._iprot.trans
         tries_left = self.retries
-        last_http_exception = None
+        last_exception = None
+        open_finished = False
         while tries_left > 0:
             try:
                 log.debug('Attempting to open transport (tries_left=%s)',
                           tries_left)
                 open_transport(transport)
+                open_finished = True
                 log.debug('Transport opened')
                 func = getattr(self.client, func_name)
                 return func(request)
-            except socket.error:
-                log.exception('Failed to open transport (tries_left=%s)',
-                              tries_left)
-                last_http_exception = None
-            except TTransportException:
-                log.exception('Failed to open transport (tries_left=%s)',
-                              tries_left)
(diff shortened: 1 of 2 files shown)


impala/tests/test_dbapi_connect.py
```diff
@@ -252,6 +252,44 @@ def test_https_connection(self):
...
 
+    def test_retry_dml(self):
+        """Regression test for #549."""
+        # Connection with higher timeout to avoid errors.
+        self.connection = connect(ENV.host, ENV.port, timeout=10)
+        # Connect with low timeout to trigger failed RPCs.
+        # TODO: This reproduces the issue in the default Impala dev env (no local catalog),
+        #       but may fail to reproduce it in the future if Impala will get faster.
+        #       Ideally the error would be injected in a more stable way.
+        low_timeout_connection = connect(ENV.host, ENV.port, timeout=3)
```
rizaon, line 263: Create a constant for both short and long timeout value.

joemcdonnell:
I think this fix makes sense.

It looks like impala-shell uses entirely different retry logic, and it limits retries to HTTP clients: https://github.com/apache/impala/blob/master/shell/impala_client.py#L692-L696

impala/hiveserver2.py
```diff
@@ -1144,41 +1144,45 @@ def __init__(self, client, retries=3):
...
-                log.exception('Failed to open transport (tries_left=%s)',
-                              tries_left)
-                last_http_exception = None
-            except TTransportException:
-                log.exception('Failed to open transport (tries_left=%s)',
-                              tries_left)
-                last_http_exception = None
+            except socket.error as e:
+                if open_finished and not safe_to_retry: raise e
+                msg = "RPC failed" if open_finished else "Failed to open transport"
```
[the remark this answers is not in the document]
csringhofer: Instead of creating a nested function merged the expect blocks,

csringhofer:
> I think this fix makes sense.
> 
> It looks like impala-shell uses entirely different retry logic, and it limits retries to HTTP clients: https://github.com/apache/impala/blob/master/shell/impala_client.py#L692-L696

Looked a bit into whether the non-http retries can be actually useful.
TLDR: for non-SSL sockets it seems useful, while for SSL sockets it only helps with idempotent RPCs

The code tries to ensure that the transport is working with
def open_transport(transport):
    if not transport.isOpen():
        transport.open()

When using raw socket, isOpen() peeks the socket (since Thrift 0.14.0) https://github.com/apache/thrift/blob/cb9ceada554f47aa5ebbedfe3984de0983cf0226/lib/py/src/transport/TSocket.py#L92 
So it can detect if the connection is broken and reestablish it with transport.open(). One caveat is that open() will throw TTransportException.ALREADY_OPEN in this case, as the failed isOpen() doesn't close the transport. (I will fix this in the next patch set)
Not sure if it is really useful to have retry loop in this case - if transport.open() threw an error, then retrying it without any sleep will probably also fail.

In case of SSLSocket currently isOpen won't check the underlying socket, see 
https://issues.apache.org/jira/browse/THRIFT-5595 . This was improved in Thrift 0.17.0 (Impyla still uses 0.16.0), but my impression is that it can still return false positive, so in that case that error will be detected during the actual Thrift RPC.

If there is no reliable way to peek SSLSockets, then I think that a sane workaround would be to do a cheap RPC without side-effect before doing a non-idempotent RPC. The time since the last RPC could be checked, e.g. only ping impalad if there was no RPC in the last few seconds.

impala/tests/test_dbapi_connect.py
```diff
@@ -236,22 +238,62 @@ def test_jwt_auth_with_ldap_password(self):
...
+        cur.execute(create)
+        successful_inserts = 0
+        NUM_INSERTS = 5
+        for i in range(NUM_INSERTS):
+            try:
+                cur.execute(invalidate)
+                cur.execute(insert)
+                successful_inserts += 1
+            except:
+                con = connect(ENV.host, ENV.port, timeout=3)
```
rizaon, line 283: Replace `timeout=3` with `LOW_TIMEOUT_S`

impala/tests/test_dbapi_connect.py
```diff
@@ -236,22 +238,62 @@ def test_jwt_auth_with_ldap_password(self):
...
 
+    def test_retry_dml(self):
+        """Regression test for #549."""
+        # Connection with higher timeout to avoid errors.
+
+        self.connection = connect(ENV.host, ENV.port, timeout=TIMEOUT_S)
+        # Connect with low timeout to trigger failed RPCs.
+        # TODO: This reproduces the issue in the default Impala dev env (no local catalog),
+        #       but may fail to reproduce it in the future if Impala will get faster.
+        #       Ideally the error would be injected in a more stable way.
```
rizaon, line 265: Just curious, can you utilize `debug_action` option to inject fault?

impala/hiveserver2.py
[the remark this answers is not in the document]
rizaon: ack.

impala/tests/test_dbapi_connect.py
rizaon: Otherwise, code looks OK.

rizaon (approved):

csringhofer:
Thanks for the reviews!

PR #551, opened 2024-09-05
Subtract RPC time from sleep in _wait_to_finish

csringhofer:
The goal is to support "long poll" (IMPALA-13294). When query option
long_polling_time_ms is set, the impala server will wait in
GetOperationStatus for this time (or until the query status changes).
This allows detecting earlier that a query has finished without making
GetOperationStatus RPCs more frequent.

If long_polling_time_ms is not used then the effect should be minor -
GetOperationStatus is quick RPC so the time it takes should mainly
come from network delay.

_get_sleep_interval() is not changed (min 0.01s, max 1s) to avoid
regression in existing use cases. It could be useful to override
this in a later patch based on the value of long_polling_time_ms.

### DIFF
--- a/impala/hiveserver2.py
+++ b/impala/hiveserver2.py
@@ -471,6 +471,7 @@ class HiveServer2Cursor(Cursor):
             return
         loop_start = time.time()
         while True:
+            start_rpc_time = time.time()
             req = TGetOperationStatusReq(operationHandle=self._last_operation.handle)
             resp = self._last_operation._rpc('GetOperationStatus', req, True)
             self._last_operation.update_has_result_set(resp)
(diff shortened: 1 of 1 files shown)


joemcdonnell (commented):
The code looks good to me. Thanks for putting this together.

I think we should give some more detail in the commit message. Just a couple sentences about why this works both with and without long polling.

csringhofer:
> The code looks good to me. Thanks for putting this together.
> 
> I think we should give some more detail in the commit message. Just a couple sentences about why this works both with and without long polling.

Added more info to the commit message.

joemcdonnell (approved):
Thanks, this looks good!

Issue #557, opened 2024-11-05
Allow users to add custom http headers to outgoing messages when using hs2-http


bartash:
In a modern Impala deployment hs2-http protocol is used in a system where http messages pass through one or more http proxies. Some of these proxies add their own http message headers to messages as they are forwarded. It would be useful to test Impala with some of the message headers that are added by http proxies. In particular the case where there are multiple http headers with the same name is hard to simulate with clients such as Impyla or Impala Shell. This is partly because these clients store http headers in a Python dict which does not allow duplicate keys.

PR #558, opened 2024-11-05
Allow users to add custom http headers when using hs2-http (#557)

bartash:
In a modern Impala deployment hs2-http protocol is used in a system where http messages pass through one or more http proxies. Some of these proxies add their own http message headers to messages as they are forwarded. It would be useful to test Impala with some of the message headers that are added by http proxies. In particular the case where there are multiple http headers with the same name is hard to simulate with clients such as Impyla or Impala Shell. This is partly because these clients store http headers in a Python dict which does not allow duplicate keys.

Extend the Impyla connect() method to add
a 'get_user_custom_headers_func' parameter. This specifies a function that is called as http message headers are being written. The function should return a list of tuples, each tuple containing a key-value pair. This allows duplicate headers to be set on outgoing messages.

TESTING
Add test code which implements a reverse http proxy, which allows test code to access the outgoing http message headers generated by Impyla. Add a test using this proxy which validates the new feature.

The new test code requires a new python package 'requests'. I think there is not away to add this requirement automatically so I added a note to README.md

All tests pass on Python2 and Python3.

Fix TestHS2FaultInjection to  use setup_method() and teardown_method() so as to work in Python3

### DIFF
--- a/impala/_thrift_api.py
+++ b/impala/_thrift_api.py
@@ -68,12 +68,14 @@ class ImpalaHttpClient(TTransportBase):
   MIN_REQUEST_SIZE_FOR_EXPECT = 1024
 
   def __init__(self, uri_or_host, port=None, path=None, cafile=None, cert_file=None,
-               key_file=None, ssl_context=None, http_cookie_names=None):
+               key_file=None, ssl_context=None, http_cookie_names=None,
+               get_user_custom_headers_func=None):
     """ImpalaHttpClient supports two different types of construction:
 
     ImpalaHttpClient(host, port, path) - deprecated
     ImpalaHttpClient(uri, [port=<n>, path=<s>, cafile=<filename>, cert_file=<filename>,
-        key_file=<filename>, ssl_context=<context>, http_cookie_names=<cookienamelist>])
+        key_file=<filename>, ssl_context=<context>, http_cookie_names=<cookienamelist>],
+        get_user_custom_headers_func=<function_setting_http_headers>)
 
     Only the second supports https.  To properly authenticate against the server,
     provide the client's identity by specifying cert_file and key_file.  To properly
@@ -85,6 +87,10 @@ class ImpalaHttpClient(TTransportBase):
     one of these names is returned in an http response by the server or an intermediate
     proxy then it will be included in each subsequent request for the same connection. If
     it is set as wildcards, all cookies in an http response will be preserved.
+    The optional get_user_custom_headers_func parameter can be used to add http headers
+    to outgoing http messages when using hs2-http protocol. The parameter should be a
+    function returning a list of tuples, each tuple containing a key-value pair
+    representing the header name and value.
     """
     if port is not None:
       warnings.warn(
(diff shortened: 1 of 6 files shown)


impala/tests/test_http_connect.py
```diff
@@ -66,10 +67,85 @@ def __init__(self):
...
+      SimpleHTTPServer.SimpleHTTPRequestHandler.__init__(self, request, client_address,
+                                                    server)
+
+    def do_POST(self):
+      # Read the body of the incoming http post message.
+      data_string = self.rfile.read(int(self.headers['Content-Length']))
+      # Save the http headers from the message in a class variable.
+      RequestHandlerProxy.saved_headers = self.decode_raw_headers()
+      # Forward the http post message to Impala and get a response message.
+      response = requests.post(url="http://localhost:28000/cliservice",
```
csringhofer, line 97: The port should be configurable with env var IMPYLA_TEST_HTTP_PORT

impala/tests/test_http_connect.py
```diff
@@ -66,10 +67,85 @@ def __init__(self):
...
+      for key, value in response.headers.items():
+        self.send_header(keyword=key, value=value)
+      self.end_headers()
+      # Send the message body.
+      self.wfile.write(response.content)
+      self.wfile.close()
+
+    def decode_raw_headers(self):
+      """Decode a list of header strings into a list of tuples, each tuple containing a
+      key-value pair. The details of how to get the headers are differs between Python2
```
csringhofer, line 113: nit: headers are differs -> are different/differs?

impala/_thrift_api.py
```diff
@@ -220,12 +232,24 @@ def setCustomUserAgent(self, user_agent):
   def setGetCustomHeadersFunc(self, func):
     self.__get_custom_headers_func = func
 
-  # Update HTTP headers based on the saved cookies and auth mechanism.
+  # Set the user-defined callback function which adds custom HTTP headers to outgoing
+  # messages.
+  def setGetUserDefinedCustomHeadersFunc(self, func):
```
csringhofer, line 237: Is this used anywhere?
Also, couldn't the test use this function to set the callback so that passing get_user_custom_headers_func through constructors can be avoided?

impala/tests/test_http_connect.py
bartash: Fixed, thanks

impala/tests/test_http_connect.py
bartash: Done, thanks

impala/_thrift_api.py
bartash: Not used and removed.
I agree that it is ugly to have to change the constructors. But the ImpalaHttpClient is hidden away so it is hard to call methods on it, especially before connect() is called. And connect() sends messages which we might want to know about. 
I welcome suggestions for better structure.

bartash:
@csringhofer please can you take another look when you get a chance

PR #562, opened 2025-02-02
Fix proxy-authentication headers for Python 3.* and long basic credential encodings

### LINKED ISSUE #517: :bug: Bug in proxy setting
https://github.com/cloudera/impyla/blob/c077bedb142cebdd266148b37f9888d04f730d14/impala/_thrift_api.py#L163C7-L163C38

This seems to be a bug

```python
def basic_proxy_auth_header(proxy):
    if proxy is None or not proxy.username:
      return None
    ap = "%s:%s" % (urllib.parse.unquote(proxy.username),
                    urllib.parse.unquote(proxy.password))
    cr = base64.b64encode(ap).strip()
    return "Basic " + cr
```

`ap` here is a string hence `base64.b64encode(ap)` fails as it expects bytes (e.g. `ap.encode()`?)

Also not sure about the purpose of the strip on a byte object - it has no effect.

### LINKED ISSUE #540: PLAIN auth with http transport fails producing valid BASIC auth header for long username/password combinations
Impyla version: v0.18.0
Python version: 3.10.9

### Minimal Example:

``` 

con = connect(
    "https://my-impala-host",
    auth_mechanism="PLAIN",
    use_http_transport=True,
    user="foo",
    password="this-is-a-very-long-password-that-is-really-very-long-and-goes-even-longer",
)

con.cursor()
``` 

Produces `ValueError: Invalid header value b'Basic Zm9vOnRoaXMtaXMtYS12ZXJ5LWxvbmctcGFzc3dvcmQtdGhhdC1pcy1yZWFsbHktdmVyeS1sb25n\nLWFuZC1nb2VzLWV2ZW4tbG9uZ2Vy'`

### Expected Behavior:

Long password and user tuples are gracefully handled, either by setting limits or by robustly producing Basic auth headers also for long user/password combinations

### Where is this problem coming from?

[`base64.encodebytes`](https://docs.python.org/3/library/base64.html#base64.encodebytes) inserts newlines after every 76 bytes of output. These would need to be filtered out [here](https://github.com/cloudera/impyla/blob/acbd481dde28d85976dfc777f888b32ad6c8d721/impala/_thrift_api.py#L422-L426)

Alternatively, use `base64.b64encode`

paulmayer:
- Added tests that cover the descriptions of #540 and #517 
- Changed plain/ldap auth and proxy authentication to use the same utility function to encode credentials

Fixes #540
Fixes #517 

Note:
1. Tested with 2.7 and 3.7+
2. Tried to use module-local conventions (e.g. unittest/pytest)
3. I'm using impala quickstarter (4.4.1) for local test execution, no hive tests executed
4. `impala.tests.test_dbapi_connect.ImpalaConnectionTests.test_retry_dml` generally fails for me (and failed also before making any modifcations)

### DIFF
--- a/impala/_thrift_api.py
+++ b/impala/_thrift_api.py
@@ -37,6 +37,7 @@ import ssl
 import sys
 
 from impala.error import HttpError
+from impala.util import get_basic_credentials_for_request_headers
 from impala.util import get_logger_and_init_null
 from impala.util import get_all_matching_cookies, get_all_cookies, get_cookie_expiry
 
@@ -177,10 +178,10 @@ class ImpalaHttpClient(TTransportBase):
   def basic_proxy_auth_header(proxy):
     if proxy is None or not proxy.username:
       return None
-    ap = "%s:%s" % (urllib.parse.unquote(proxy.username),
-                    urllib.parse.unquote(proxy.password))
-    cr = base64.b64encode(ap).strip()
-    return "Basic " + cr
+    return "Basic " + get_basic_credentials_for_request_headers(
+       user=urllib.parse.unquote(proxy.username),
+       password=urllib.parse.unquote(proxy.password),
+    )
 
   def using_proxy(self):
     return self.realhost is not None
@@ -462,11 +463,7 @@ def get_http_transport(host, port, http_path, timeout=None, use_ssl=False,
             log.debug('get_http_transport: password=fuggetaboutit')
         auth_mechanism = 'PLAIN'  # sasl doesn't know mechanism LDAP
         # Set the BASIC auth header
-        user_password = '%s:%s'.encode() % (user.encode(), password.encode())
-        try:
-            auth = base64.encodebytes(user_password).decode().strip('\n')
-        except AttributeError:
-            auth = base64.encodestring(user_password).decode().strip('\n')
+        auth = get_basic_credentials_for_request_headers(user, password)
 
         def get_custom_headers(cookie_header, has_auth_cookie):
             custom_headers = {}
--- a/impala/tests/test_http_connect.py
+++ b/impala/tests/test_http_connect.py
@@ -203,7 +203,6 @@ class TestHttpConnect(object):
     rows = cur.fetchall()
     assert rows == [(1,)]
 
-    print(http_proxy_server.get_headers())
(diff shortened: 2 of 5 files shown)


csringhofer (approved):
Lgtm, thanks for the fix and tests!

csringhofer:
>impala.tests.test_dbapi_connect.ImpalaConnectionTests.test_retry_dml generally fails for me (and failed also before making any modifcations)

I wrote test_retry_dml in a hurry and it is not surprising that it is flaky - practically it relies on Impala's non local catalog's slowness due to statestore's low update frequency. I will try to rework it to be more reliable.

>I'm using impala quickstarter (4.4.1) for local test execution, no hive tests executed
Can you give more info on the env? What is  exactly impala quickstarter? :)

csringhofer:
@paulmayer 
Created a PR to make test_retry_dml more reliable by using a Impala debug action to trigger the timeout:
https://github.com/cloudera/impyla/pull/566
Tested it with legacy and local catalog Impala (on local catalog it was failing before the patch) and checked that allowing retries during ExecuteStatement leads to failing the test.
Can you check if it is green is your environment?

csringhofer:
@paulmayer 
I checked if the same issue is present in impala-shell, and long password are also mishandled there.
Created ticket [IMPALA-13746](https://issues.apache.org/jira/browse/IMPALA-13746)

Are you interested in also implementing this for impala-shell, or it is ok if someone else ports your solution there?
impala-shell resides in the of Apache Impala git repo, otherwise the Python code is very similar to impyla. Note that Impala usess gerrit for new changes instead of github PRs. See https://cwiki.apache.org/confluence/display/IMPALA/Contributing+to+Impala for more info on how to contribute to the impala repository.

paulmayer:
@csringhofer happy to add it once i'm through with the registration process, but given it's a one-line change of https://github.com/paulmayer/impala/blob/master/shell/impala_client.py#L588-L592 to ` auth=base64.b64encode(user_passwd.encode()).decode().strip()`, I'm of course OK if it's done by then :)

csringhofer:
@paulmayer  this became important for impala-shell so I uploaded a patch and referred you in the commit message
https://gerrit.cloudera.org/#/c/22780/

PR #564, opened 2025-02-04
Bugfix/retry configuration not passed to rpc operations

### LINKED ISSUE #563: HS2Service retries do not propagate down to RPC operations
Neither [HS2Service.open_session](https://github.com/cloudera/impyla/blob/master/impala/hiveserver2.py#L1278) nor [HS2Session._get_operation](https://github.com/cloudera/impyla/blob/master/impala/hiveserver2.py#L1389) pass retry configuration, which means that operations like ` TGetOperationStatusReq` fail irrespective of connection retry configurations after 3 attempts.

paulmayer:
Pass through `retries`. Adapted tests to check whether downstream rpc operations use configured retry amounts.

Fixes #563

### DIFF
--- a/impala/hiveserver2.py
+++ b/impala/hiveserver2.py
@@ -1286,7 +1286,8 @@ class HS2Service(ThriftRPC):
         resp = self._rpc('OpenSession', req, True)
         return HS2Session(self, resp.sessionHandle,
                           resp.configuration,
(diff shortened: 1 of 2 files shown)


csringhofer (approved):
Thanks for the fix!
Ran the tests locally and they passed.

Issue #570, opened 2025-02-23
Inconsistent bracket handling of ipv6 addresses


csringhofer:
http and binary connections expect different format for ipv6 addresses:

these work:
impala.dbapi.connect("::1", 21050)
impala.dbapi.connect("[::1]", 28000, use_http_transport=True)
these throw exceptions:
impala.dbapi.connect("[::1]", 21050)
impala.dbapi.connect("::1", 28000, use_http_transport=True)

I think that ideally both formats would be accepted.

PR #571, opened 2025-03-06
Fix IPv6 address handling in hs2-http2 (#570)

csringhofer:
- add brackets to ipv6 addresses in urls (e.g. https://[::1]:28000)
- fix tests to work with IMPYLA_TEST_HOST=::1

Tested with WIP patch for IPv6 support in Impala:
https://gerrit.cloudera.org/#/c/22527/

### DIFF
--- a/impala/tests/test_http_connect.py
+++ b/impala/tests/test_http_connect.py
@@ -24,9 +24,19 @@ from six.moves import http_client
 from six.moves import socketserver
 
 from impala.error import HttpError
-from impala.tests.util import ImpylaTestEnv
+from impala.tests.util import ImpylaTestEnv, is_ipv6_only_host
 
 ENV = ImpylaTestEnv()
+
+IS_IPV6_ONLY_HOST = is_ipv6_only_host(ENV.host, ENV.port)
+LOCAL_HOST = "::1" if IS_IPV6_ONLY_HOST else "127.0.0.1"
+
+# socketserver.TCPServer cannot listen both on ipv4 and ipv6. Listen to ipv6
+# if the hs2-http server has only ipv6 address.
+class IPv4or6TcpServer(socketserver.TCPServer):
+  address_family = socket.AF_INET6 if IS_IPV6_ONLY_HOST else socket.AF_INET
+  def __init__(self, host_port, req_handler):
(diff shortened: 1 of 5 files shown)


impala/tests/test_http_connect.py
```diff
@@ -24,10 +24,20 @@
 from six.moves import socketserver
 
 from impala.error import HttpError
-from impala.tests.util import ImpylaTestEnv
+from impala.tests.util import ImpylaTestEnv, is_ipv6_only_host
 
 ENV = ImpylaTestEnv()
 
+IS_IP6_ONLY_HOST = is_ipv6_only_host(ENV.host, ENV.port)
```
gaallas, line 31: Is there a specific reason why IS_IP6_ONLY_HOST is spelled with IP6_ONLY instead of IPV6_ONLY?
I'm not saying it's wrong, but IPv6 is spelled IPv6 everywhere else, so it definitely made me wonder about the difference here.

csringhofer: I agree, change it to IS_IPV6_ONLY_HOST

impala/tests/test_http_connect.py
```diff
@@ -24,10 +24,20 @@
...
 
 ENV = ImpylaTestEnv()
 
+IS_IPV6_ONLY_HOST = is_ipv6_only_host(ENV.host, ENV.port)
+LOCAL_HOST = "::1" if IS_IPV6_ONLY_HOST else "127.0.0.1"
+
+# socketserver.TCPServer cannot listen both on ipv4 and ipv6. Listen to ipv6
+# if the hs2-http server has only ipv6 address.
+class IPv4or6TcpServer(socketserver.TCPServer):
+  address_family = socket.AF_INET6 if IS_IPV6_ONLY_HOST else socket.AF_INET
```
MikaelSmith, line 37: This seems weird, but it's what the [docs](https://docs.python.org/3/library/socketserver.html) say to do.

csringhofer: there is some interesting discussion about this including Guido himself - it seems that this is very old code that they don't want to change/improve:
https://discuss.python.org/t/ipv6-in-socketserver/79551/8

PR #572, opened 2025-03-06
Remove versioneer and add Python3.12/3.13 testing

csringhofer:
- remove versioneer and set version manually
- add python 3.12/3.13 to tox.ini

versioneer has no version which supports both Python 2.7 and >=3.12 Though python2.7 is EOL, Apache Impala tests still use it, making its removal problematic (see #532, #533). Once Python 2.7 support is removed versioneer or some other similar tool can be added again.

Testing:
- ran tests with supported Python versions (including 3.12 and 3.13)


impala/__init__.py
```diff
@@ -14,6 +14,4 @@
 
 from __future__ import absolute_import
 
-from ._version import get_versions
-__version__ = get_versions()['version']
-del get_versions
+__version__ = u'v0.20.0'
```
MikaelSmith, line 17: Should this match the version in setup.py?

csringhofer: Ouch, yes, probably the main reason to use versioneer is to avoid these kind of mistakes :)
https://packaging.python.org/en/latest/discussions/single-source-version/

This could be trouble if it was in an inconsistent state when a release is created, though generally __version__ is not used when a library is imported.

csringhofer: btw fixed it to use v0.20.0 is both cases
will have to bump both when preparing a release

MikaelSmith: Let's add a code comment to remind people to update the other location as well. For both spots.

csringhofer: good idea, done

MikaelSmith (approved):

PR #578, opened 2025-03-11
Update build_summary_table function

rizaon:
This patch update build_summary_table to match the same function in impala-shell.
https://github.com/apache/impala/blob/a07bf84/shell/impala_client.py#L113

Testing:
Run and pass following command
```
tox -- -ktest_build_summary_table
```

### DIFF
--- a/impala/hiveserver2.py
+++ b/impala/hiveserver2.py
@@ -29,6 +29,7 @@ from six.moves import range
 from thrift.transport.TTransport import TTransportException
 from thrift.Thrift import TApplicationException
 from thrift.protocol.TBinaryProtocol import TBinaryProtocolAccelerated
+from impala._thrift_gen.ExecStats.ttypes import TExecStats
 from impala._thrift_gen.TCLIService.ttypes import (
     TOpenSessionReq, TFetchResultsReq, TCloseSessionReq,
     TExecuteStatementReq, TGetInfoReq, TGetInfoType, TTypeId,
@@ -727,8 +728,9 @@ class HiveServer2Cursor(Cursor):
 
     def build_summary_table(self, summary, output, idx=0,
                             is_fragment_root=False, indent_level=0):
-        return build_summary_table(summary, idx, is_fragment_root,
-                                   indent_level, output)
+        return build_exec_summary_table(
+            summary, idx, indent_level, is_fragment_root, output, is_prettyprint=True,
+            separate_prefix_column=False)
 
     def get_databases(self):
         def op():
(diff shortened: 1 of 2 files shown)


impala/hiveserver2.py
```diff
@@ -1585,23 +1593,27 @@ def build_summary_table(summary, idx, is_fragment_root, indent_level, output):
...
-    if len(node.exec_stats) > 0:
+    if node.exec_stats is not None:
+        for stats in node.exec_stats:
+            for attr in attrs:
+                val = getattr(stats, attr)
+                if val is not None:
+                    setattr(agg_stats, attr, getattr(agg_stats, attr) + val)
+                    setattr(max_stats, attr, max(getattr(max_stats, attr), val))
+
+    if node.exec_stats is not None and node.exec_stats:
```
csringhofer, line 1604: wouldn't be simply if node.exec_stats: enough?

impala/hiveserver2.py
```diff
@@ -1634,22 +1647,49 @@ def prettyprint_units(unit_val):
...
-           node.label_detail]
+    instances = 0
+    if node.exec_stats is not None:
+        instances = len(node.exec_stats)
+    latency = max_stats.latency_ns
+    cardinality_est = est_stats.cardinality
+    memory_used = max_stats.memory_used
+    memory_est = est_stats.memory_used
+    if (is_prettyprint):
+        avg_time = prettyprint_time(avg_time)
```
csringhofer, line 1658: probably it would be better to do it in another commit, but I think that formatting could be move to another file logic, e.g. exec_summary.py

impala/tests/test_impala.py
```diff
@@ -112,3 +112,69 @@ def validate_log(cur):
...
+        ['02:HASH JOIN', 1, 1, '0', '0', '1.94 MB', 'INNER JOIN, BROADCAST'],
+        ['|--03:EXCHANGE', 1, 1, '0', '0', '16.00 KB', 'BROADCAST'],
+        ['|  F01:EXCHANGE SENDER', 1, 1, '', '', '32.00 KB', ''],
+        ['|  01:SCAN HDFS', 1, 1, '0', '0', '0 B',
+         '{0}.{1} b'.format(tmp_db_lower, empty_table)],
+        ['00:SCAN HDFS', 1, 1, '0', '0', '0 B',
+         '{0}.{1} a'.format(tmp_db_lower, empty_table)],
+    ]
+    validate_summary_table(output_dop_0, expected_dop_0)
+    cur.close_operation()
```
csringhofer, line 158: cur.close_operation() is not needed, execute()  (and the cur fixture after the test is finished) will call it

rizaon: Done

impala/hiveserver2.py
rizaon: Replaced this with a new exec_summary.py from https://github.com/apache/impala/commit/e73e2d40da115ed3804ffaecc0850c853b0e6330

csringhofer (approved):

PR #588, opened 2025-06-26
Unpin bitarray dependency on Python 3

csringhofer:
Python 2.7 needs bitarray < 3, but on Python bitarray 3.* can be used. The bitarray functionality used in impyla is minimal and didn't change with the major version bump.


rizaon (approved):
This looks good to me.

PR #589, opened 2025-06-26
Change cur fixture to function level

csringhofer:
test_build_summary_table failed based on what other tests ran in the same session due to using the same session level cursor which could have altered query options from other tests. The patch switches to function level cursor to isolate tests and creates new fixture session_cur to use in session/module level fixtures that need a cursor.


rizaon (approved):
Run `tox` myself and it pass.

PR #597, opened 2026-05-05
resilient columnar batch handling in hiveserver2.py

### LINKED ISSUE #596: fetchcolumnar() unconditionally breaks on empty batch, ignoring hasMoreRows, causing truncated results
## Describe the bug

`HiveServer2Cursor.fetchcolumnar()` breaks out of the fetch loop the moment it receives an empty batch, without checking whether the server signalled that more rows are still expected:

```python
# hiveserver2.py — fetchcolumnar() — current upstream (buggy)
while True:
    batch = (self._last_operation.fetch(...))
    if len(batch) == 0:
        break          # ← unconditional; ignores batch.expect_more_rows
    batches.append(batch)
```

When Impala's `FETCH_ROWS_TIMEOUT_MS` (default: 10000 ms) expires before the next batch of rows is ready, Impala returns an empty `TRowSet` with `hasMoreRows=True`. This is valid, documented server behaviour for non-blocking fetch semantics. `fetchcolumnar()` treats this empty interim response as end-of-results and stops fetching, silently discarding all remaining rows.

Both `CBatch` and `RBatch` already carry an `expect_more_rows` attribute populated from `resp.hasMoreRows`, and `_ensure_buffer_is_filled()` (the non-columnar path) already handles this correctly:

```python
# _ensure_buffer_is_filled() — correct handling of empty interim batch
self._buffer = self._last_operation.fetch(...)
if not self._buffer.expect_more_rows:
    raise StopIteration
# otherwise: loop continues and fetches again
```

`fetchcolumnar()` should do the same.

## Symptoms

- Result row counts are non-deterministic across runs of the same query.
- Counts are always a multiple of 1024 (Impala's default batch size).
- The query continues running server-side after the client has returned (visible in Cloudera Navigator / Impala query history as queries running until idle session timeout).
- No exception is raised; data is silently truncated.

## Steps to reproduce

1. Connect to an Impala cluster with `FETCH_ROWS_TIMEOUT_MS > 0` (default is 10000).
2. Run a query returning a large result set (e.g., >10 000 rows).
3. Fetch via `cursor.fetchall(columnar=True)` or any code path that calls `fetchcolumnar()` (e.g., ibis's `fetch_from_cursor`).
4. Observe that the returned row count varies across runs and is less than the actual result size.

## Proposed fix

```python
def fetchcolumnar(self):
    """Executes a fetchall operation returning a list of CBatches"""
    self._wait_to_finish()
    if not self._last_operation.is_columnar:
        raise NotSupportedError("Server does not support columnar fetching")
    batches = []
    while True:
        batch = (self._last_operation.fetch(
                     self.description,
                     self.buffersize,
                     convert_types=self.convert_types,
                     convert_strings_to_unicode=self.convert_strings_to_unicode))
        if len(batch) == 0:
            if not batch.expect_more_rows:
                break
            continue  # empty interim batch; server has more rows — keep fetching
        batches.append(batch)
    return batches
```

## Environment

- impyla version: 0.21.0
- Python: 3.10
- Impala: CDH (Cloudera)
- `FETCH_ROWS_TIMEOUT_MS`: 10000 (server default)
- `SPOOL_QUERY_RESULTS`: 1

## Mitigation / Alternative workaround

Setting `FETCH_ROWS_TIMEOUT_MS=0` at session open disables non-blocking fetch, preventing Impala from returning empty interim batches. This avoids the bug but is not a fix.

## Related

- IMPALA-10566 — requests that `FETCH_ROWS_TIMEOUT_MS` default be changed to 0; this issue shows that impyla has a client-side bug that makes the current default unsafe regardless of whether that Jira is resolved.

shatteringlass:
fixes #596

### DIFF
--- a/impala/tests/test_impala.py
+++ b/impala/tests/test_impala.py
@@ -67,7 +67,7 @@ def test_has_more_rows_fetchcolumnar(cur, bigger_table):
                    from {0}
                    where s != cast(sleep(2) as string)""".format(bigger_table))
     expected_rows = [("row{0}".format(i),) for i in xrange(BIGGER_TABLE_NUM_ROWS)]
-    assert sorted(cur.fetchcolumnar()) == sorted(expected_rows)
+    assert sorted(cur.fetchcolumnar()[0]) == sorted(expected_rows)
 
 
 @fixture(scope='function')
--- a/impala/tests/test_impala.py
+++ b/impala/tests/test_impala.py
@@ -53,6 +53,23 @@ def test_has_more_rows(cur, bigger_table):
     expected_rows = [("row{0}".format(i),) for i in xrange(BIGGER_TABLE_NUM_ROWS)]
     assert sorted(cur.fetchall()) == sorted(expected_rows)
 
+
+def test_has_more_rows_fetchcolumnar(cur, bigger_table):
+    """Test that impyla correctly handles empty row batches returned with the
+    hasMoreRows flag when using columnar batching."""
+    # Set the fetch timeout very low and add sleeps so that Impala will return
+    # empty batches. Run on a single node with a single thread to make as predictable
+    # as possible.
+    cur.execute("set fetch_rows_timeout_ms=1")
+    cur.execute("set num_nodes=1")
+    cur.execute("set mt_dop=1")
+    cur.execute("""select *
+                   from {0}
+                   where s != cast(sleep(2) as string)""".format(bigger_table))
+    expected_rows = [("row{0}".format(i),) for i in xrange(BIGGER_TABLE_NUM_ROWS)]
+    assert sorted(cur.fetchcolumnar()) == sorted(expected_rows)
+
+
 @fixture(scope='function')
 def empty_table(cur):
     table_name = 'tmp_empty_table'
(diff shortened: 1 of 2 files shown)


csringhofer:
Thanks for fixing this!
Fyi I am trying to create an alpha release 0.23a1 for one specific change, but once I am done, I will be merge this.

Meanwhile can you try adding tests? https://github.com/cloudera/impyla/blob/a0da5bf4b6f731f400e25be279b8b083ce5ae3c8/impala/tests/test_impala.py#L41 has an example for exercising expect_more_rows.
If running the tests is a problem, I can also add it in a follow up commit.

shatteringlass:
Hi @csringhofer, I have added a test that should address your concern but I am unable to run it at the moment.

impala/tests/test_impala.py
```diff
@@ -53,6 +53,23 @@ def test_has_more_rows(cur, bigger_table):
...
+    # empty batches. Run on a single node with a single thread to make as predictable
+    # as possible.
+    cur.execute("set fetch_rows_timeout_ms=1")
+    cur.execute("set num_nodes=1")
+    cur.execute("set mt_dop=1")
+    cur.execute("""select *
+                   from {0}
+                   where s != cast(sleep(2) as string)""".format(bigger_table))
+    expected_rows = [("row{0}".format(i),) for i in xrange(BIGGER_TABLE_NUM_ROWS)]
+    assert sorted(cur.fetchcolumnar()) == sorted(expected_rows)
```
csringhofer: This didn't pass - [0] is needed after fetchcolumnar()
With that added I checked that without the fix it fails.

csringhofer (approved):
Thanks, the test passes now

PR #601, opened 2026-05-11
Allow verifying server using default CA certificates

csringhofer:
Add verify_cert arg to connect() to verify the server when use_ssl=True but ca_cart is not. In this case if verify_cert=True then Python will validate the server using default CA certs.

For details of default CA handling in Python see:
https://docs.python.org/3/library/ssl.html

Testing:
- added negative tests for connecting to Impala with self signed cert
- no positive tests - these would need a proper certificate or forcing Python to accept a self-signed certificate

### DIFF
--- a/impala/_thrift_api.py
+++ b/impala/_thrift_api.py
@@ -382,10 +382,10 @@ class ImpalaHttpClient(TTransportBase):
       raise HttpError(self.code, self.message, body, self.headers)
 
 
-def get_socket(host, port, use_ssl, ca_cert):
+def get_socket(host, port, use_ssl, ca_cert, verify_cert):
     # based on the Impala shell impl
-    log.debug('get_socket: host=%s port=%s use_ssl=%s ca_cert=%s',
-              host, port, use_ssl, ca_cert)
+    log.debug('get_socket: host=%s port=%s use_ssl=%s ca_cert=%s verify_cert=%s',
+              host, port, use_ssl, ca_cert, verify_cert)
 
     if use_ssl:
         from thrift.transport.TSSLSocket import TSSLSocket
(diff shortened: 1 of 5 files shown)


impala/tests/test_dbapi_connect.py
```diff
@@ -256,7 +256,21 @@ def test_ssl_connection_with_cert(self):
     def test_ssl_connection_wrong_cert(self):
         try:
             connect(
-                ENV.host, ENV.port, use_ssl=True, timeout=TIMEOUT_S, ca_cert=ENV.ssl_wrong_cert)
```
MikaelSmith, line 259: Did you mean to modify this test? Looks like it now duplicates the next test.

impala/_thrift_api.py
```diff
@@ -409,14 +410,15 @@ def get_http_transport(host, port, http_path, timeout=None, use_ssl=False,
...
+                       get_user_custom_headers_func=None, verify_cert=False):
     host_url = "[%s]" % host if ":" in host else host # add brackets for ipv6 address
     # TODO: support timeout
     if timeout is not None:
         log.error('get_http_transport does not support a timeout')
     if use_ssl:
         ssl_ctx = ssl.create_default_context(cafile=ca_cert)
-        if ca_cert:
+        if ca_cert or verify_cert:
+          ssl_ctx.check_hostname = True
```
MikaelSmith, line 421: This looks like it was missed, it should have a separate test.

impala/tests/test_dbapi_connect.py
csringhofer: Thanks for spotting, this was a copy paste mistake.

MikaelSmith (approved):

impala/_thrift_api.py
MikaelSmith: Confirmed in recent Pythons - https://github.com/python/cpython/blob/v3.12.13/Lib/ssl.py#L698 and https://github.com/python/cpython/blob/v3.9.25/Lib/ssl.py#L744 - this was already the default.

Issue #471, opened 2021-09-18
Undeclared dependency on puresasl


cpcloud:
At some point `impyla` started depending on `puresasl` (https://github.com/cloudera/impyla/blob/master/impala/sasl_compat.py#L1), but didn't declare the dependency the `setup.py`. Unsurprisingly this is causing downstream environments to have an insufficient dependency graph to be able to use impyla:  https://github.com/ibis-project/ibis/runs/3632559971

csringhofer:
+1, added it as direct dependency in https://github.com/cloudera/impyla/pull/603

PR #603, opened 2026-05-13
Add direct dependency to pure-sasl (#471)

csringhofer:
impyla uses pure-sasl directly but it was not mentioned in setup.py. Installs still worked due to inderectly depending on it through thrift_sasl. Added the same dependency as the one in thrift_sasl.


gaallas (changes requested):
Nit: typo in commit message: "inderectly"

csringhofer:
My last force push rebased the change (moving dependency from setup.py to pyproject.toml) and resolved @gaallas 's comment.

Issue #469, opened 2021-09-14
Consider dropping support for Python 2.7


cpcloud:
It looks like Python 2.7 (and even 2.6 :grimacing: ) is still listed as supported in `setup.py`. This makes downstream packaging efforts more annoying than they need to be, for a version of the language that was EOLd in 2020.

Do we have any idea how widespread impyla + python 2.7 usage is?

cpcloud:
Looks like this was already brought up in https://github.com/cloudera/impyla/issues/389

PR #606, opened 2026-05-28
Remove Python 2.7 and 3.7 support (#469)

csringhofer:
With this change impyla tests pass from Python 3.8 to 3.14. With 3.7 still most tests pass, with the exception of test_pandas_dataframe_to_sql which fails due to older pandas version.

Dependencies are unpinned in tox.ini - this means that new versions of libraries may break tests without any change in impyla, allowing catching such issues earlier. pytest is also unpinned, leading to using 8.* on Python 3.8/3.9 and 9.* on newer Pythons.


MikaelSmith (approved):
I'm a bit mixed about leaving dependencies totally unpinned, but I guess that's already been true for some. Consuming projects can still pin them if needed.

jasonmfehr (approved):

jasonmfehr:
We should also update the major version to `1.0.0` with this PR since it will break backwards compatibility.

csringhofer:
> We should also update the major version to `1.0.0` with this PR since it will break backwards compatibility.

There have been a couple of breaking changes between minors so far, and dropping support for older pythons is actually something that pip can handle well AFAIK, so download the freshest version supported with current Python.
I was also thinking about switching to 1.0.0, but mainly due to setting verify_cert=True by default.

PR #607, opened 2026-05-29
Switch to pyproject.toml

csringhofer:


tox.ini
```diff
@@ -1,7 +1,7 @@
 [tox]
 requires =
-    tox>=4
-    virtualenv<20.22.0
+    tox>=4,
+    virtualenv
```
gaallas: Does this (maybe tox? I confess ignorance here) really require the `virtualenv` package? Wouldn't Python3's standard `venv` module be sufficient? Note that this is not a change request, I would just prefer oreventing unnecessary complexity.

pyproject.toml
```diff
@@ -0,0 +1,57 @@
...
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
+    "Programming Language :: Python :: 3.14",
+]
+dependencies = [
+    "bitarray",
+    "thrift==0.16.0",
```
gaallas, line 34: Shouldn't `bitarray` be pinned too?

gaallas (commented):
LGTM, and thanks for the update (long overdue, I guess).
The comments are more inquiries than change requests.

pyproject.toml
csringhofer: pinning bitarray caused problems for some users in the past

tox.ini
csringhofer: good point, without the constraint mentioning "virtualenv" is redundant

tox uses virtualenv instead of venv AFAIK

pyproject.toml
gaallas: Ack

tox.ini
gaallas: OK; I guess we can experiment with the replacement later.

gaallas (approved):
Thanks for the updates, looks good.

PR #614, opened 2026-06-04
Add script to run all tests

csringhofer:
bin/run_tests_in_impala_dev_env.sh allow running all tests by restarting Impala cluster with different settings (ssl, jwt). Kerberos and LDAP testing are still missing as these need more complex setup.

This needed some reorganization in test_dbapi_connect to allow selecting tests by using marking.

Other changes:
- create junit report for jenkins
- added bin/bootstrap_test_env.sh to install all supported Python versions
- removed old scripts in jenkins/ as they were stale and assumed internal Cloudera environment


MikaelSmith (approved):

bin/run_tests_in_impala_dev_env.sh
```diff
@@ -0,0 +1,76 @@
+# Copyright 2026 Cloudera Inc.
```
gaallas, line 2: Please add shebang line for `bash`; the script contains bash-specific syntax, so the correct shell kind should be ensure.

impala/tests/conftest.py
gaallas, line 1: nit: update the copyright notice dates

bin/bootstrap_test_env.sh
```diff
@@ -0,0 +1,30 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Install dependencies to run tests on each supported Python version using tox.
+
+sudo apt install --yes python3-pip
```
gaallas: Nit: I'd suggest adding `-q` to all invocations of `apt` to decrease output noise.

gaallas (commented):
None of these are blockers, they are OK to be addressed by follow-up patches.

gaallas:
I removed my earlier comments about "set +e" and friends; I realized they didn't make sense.
The rest are really just suggestions, maybe except for the shebang line at the top of `bin/run_tests_in_impala_dev_env.sh`

